PR #86, opened 2014-11-20
[fix bug 933848] Add firefox version property.

koddsson:
This looks like this:

screen shot 2014-11-19 at 21 17 24
---
screen shot 2014-11-19 at 21 17 42

Regarding UI I'm wondering if we maybe want to have some sort of filtered dropdown widget like this to filter the versions or if the extremely long list of versions is acceptable:
screen shot 2014-11-19 at 21 23 43

This does not address selecting a "current version" nor the "current version - 2" criteria as I'm not quite sure either on how that would be represented in UI.

But at least 

cc/ @Osmose @glogiotatidis


glogiotatidis:
This is way too old now so I'm going to close it. Thanks for your work @koddsson.

PR #106, opened 2015-04-16
Fix snippet_id in snippet edit and document.

glogiotatidis:


mozilla-travis:
```
snippets/base/models.py:202:61: E202 whitespace before ')'

```

glogiotatidis:
@Osmose I remember us talking about this and we agreed to go with `<snippet_id>` because it was less confusing than `{snippet_id}`. Unfortunately this does not go through the XML validator and therefore one cannot save snippets with `<snippet_id>` in either body. I changed that to `[[snippet_id]]` and wrote two lines of documentation. r?

To be on the safe side I checked that no snippet is using `<snippet_id>` on production.

mozilla-travis:
```
snippets/base/models.py:202:61: E202 whitespace before ')'

```

Osmose:
Makes sense to me, r+ :D

glogiotatidis:
Thanks!

PR #109, opened 2015-04-22
[fix bug 1157141] Add user country info to the Snippets Stats Ping

glogiotatidis:


glogiotatidis:
@alexgibson r?

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -121,6 +123,8 @@
             );
         }
         else {
+            document.getElementById('snippet_set').dataset.userCountry = '';
```
alexgibson: I'm guessing it's expected that this could result in an empty query param value for `country` in the stats ping?

alexgibson:
Code looks fine to me. One minor query but otherwise this looks good r+

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Yes, expected :)

glogiotatidis:
Thanks!

PR #105, opened 2015-04-15
[bug 1129799] Multiple Countries per Snippet

glogiotatidis:

PR #110, opened 2015-04-28
[fix bug 1155574] Add ability to select multiple countries in a single J...

glogiotatidis:
...SON snippet


glogiotatidis:
This is very similar to #105 

@pmclanahan r?

pmclanahan:
r+ :+1:

PR #112, opened 2015-05-13
Explicitly declaring css height for snippet icons.

glogiotatidis:
Don't allow SVG icons get larger than 40x50px by explicitly declaring
their height.

### DIFF
--- a/snippets/base/templates/base/includes/snippet_css.html
+++ b/snippets/base/templates/base/includes/snippet_css.html
@@ -1,36 +1,37 @@
 <style type="text/css">
-#snippetContainer div.snippet {
-  display: none;
-}
-
-#snippetContainer div.snippet img.icon {
-  margin: -0.75em 1em 0 0;
-  float: left;
-}
-
-{% if client and client.startpage_version > 1 %}
-  .snippet {
-    position: relative;
-  }
-
-  .snippet .icon {
-    position: absolute;
-    right: 400px;
-    top: 5px;
-  }
-
-  .snippet p {
-    display: table-cell;
-    height: 40px;
-    padding-left: 80px;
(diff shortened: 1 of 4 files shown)


snippets/base/templates/base/includes/snippet_css.html
```diff
@@ -1,36 +1,37 @@
...
-  .snippet .icon {
-    position: absolute;
-    right: 400px;
-    top: 5px;
-  }
+   .snippet .icon {
+       position: absolute;
+       right: 400px;
+       top: 5px;
+       height: 50px;
```
glogiotatidis: This is the only non-spacing change in this file.

mozilla-travis:
```
snippets/base/models.py:329:101: E501 line too long (101 > 100 characters)

```

mozilla-travis:
```
snippets/base/models.py:329:101: E501 line too long (101 > 100 characters)

```

glogiotatidis:
@Osmose r? I don't have a Retina display so I can't test whether setting px breaks anything on Retina but since we use px in other places I guess it will be fine. I'm curious though how Firefox handles Retina displays? Does it scale everything to 110% for example? Is there a point to using SVGs?

Osmose:
Gotta fix that broken test, but r+ otherwise. Nice work!

I tested on a Retina MacBook and I believe it works in the correct way with this fix:

- If you use an SVG, it is constrained and smooth on a Retina screen.
- If you use a PNG at 40x50, it is constrained but blurry on a Retina screen.
- If you use a PNG at 80x100, it is constrained and smooth on a Retina screen.

So I think you're good!

glogiotatidis:
thanks @Osmose  :)

PR #113, opened 2015-05-25
Cache bundles for a month.

glogiotatidis:
Bundle filename is calculated based on snippet id and snippet modified date. Thus if contents change the bundle filename changes too. By instructing the browser to cache the bundle for a month we reduce the traffic transferred in the case where bundles do not change between the 4 hour intervals or for a day or two.

We still keep the `settings.SNIPPET_BUNDLE_TIMEOUT` setting of the `fetch_pregenerated_snippets` view, which means that the browser will try to get new snippets after `settings.SNIPPET_BUNDLE_TIMEOUT` and in the case that the bundle hasn't changed it will be redirected to something already cached.

### DIFF
--- a/media/bundles/.htaccess
+++ b/media/bundles/.htaccess
@@ -1,0 +1,4 @@
+ExpiresActive on
+ExpiresDefault "access plus 1 month"
+FileETag MTime
+Header set Access-Control-Allow-Origin "*"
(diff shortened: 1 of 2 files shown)


glogiotatidis:
@Osmose I'm optimizing this even more so we can turn on 4 hour fetch intervals for mobile too. r?

Osmose:
r+, nice work!

PR #114, opened 2015-07-07
 [bug 1143870] Count snippet clicks. 

glogiotatidis:


glogiotatidis:
@alexgibson r? :)

glogiotatidis:
A bit of context: The idea is to send a metric back to our servers on every link click. To do so I store the actual link in `data-href` and set re-set the stored value to `href` after I've send the metric.

alexgibson:
Testing on https://snippets.allizom.org/show/84/

Seems to work ok, but some points to consider (up to you if you feel these warrant changing the current code):

* Because the href gets changed to `click:`, I can no longer see the link target URL in the status bar.
* Opening links in new tab is kinda broken. If I control/click to get the context menu, the link gets clicked. If I command/click, I get two tabs showing the destination link.

glogiotatidis:
Updated to address review comments. 

A special case: When user right clicks on a click (or to be more precise when user opens the menu) I still count that for a click on the link, although the user may choose another action than to open a tab, e.g. `view page source`.

@alexgibson you can try the new code at https://snippets.allizom.org/show/84/

alexgibson:
Code looks good and testing on stage seems to be working well, r+ :cake:

PR #123, opened 2015-09-01
Call DOM altering functions after snippet finishes initializing.

glogiotatidis:


snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -72,20 +73,26 @@
...
-        addSnippetBlockLinks(show_snippet.getElementsByClassName('block-snippet-button'));
+        // By the time show_snippet event reaches snippet_set the snippet
+        // will have finished initializing and altering the DOM. It's the
+        // time to modifyLinks() and addSnippetBlockLinks().
+        snippet_set.addEventListener('show_snippet', function(event) {
+            var topSection = document.getElementById('topSection');
+            // Add sample rate and snippet ID to currently displayed links.
+            var parameters = ('sample_rate=' + SNIPPET_METRICS_SAMPLE_RATE + '&snippet_name=' +
+                              show_snippet_id);
+            modifyLinks(topSection.getElementsByTagName('a'), parameters);
```
schalkneethling: seeing we know we will always be in Fx only, I would do:

```
...(topSection.querySelectorAll('a'),....
```

glogiotatidis: is `querySelectorAll` a Fx only feature? Is this faster?

glogiotatidis:
@schalkneethling do you have any other comments on this? Reading your comments on https://github.com/mozilla/snippets/pull/53 if you can use the close button on the logo it means that this works as expected for you.

snippets/base/templates/base/includes/snippet_js.html
schalkneethling: No, not Fx only, all modern browsers support it and yes, unless you are doing `getElementById` it is faster.

schalkneethling:
r+ :mahjong:

schalkneethling:
I do not have big green button access so, merge at will.

glogiotatidis:
Thanks!

PR #130, opened 2015-11-12
Change Array.unshift to concat.

glogiotatidis:
Array.unshift is not supported in Fx <22. Use concat lists to mimic the
behavior.


glogiotatidis:
@schalkneethling r?

schalkneethling:
r+ Out of curiosity, do we still have to support Fx version as far back as that for about:home?

glogiotatidis:
> r+ Out of curiosity, do we still have to support Fx version as far back as that for about:home?

We don't have a support policy so atm we support as far back as we can go. I'm working on clear policy though. BTW do you know where I can find numbers of Fx usage per version?

schalkneethling:
Not of the top of my head but, I bet the Metrics folks will have a dashboard ;)

PR #131, opened 2015-11-13
Ensure that template are valid XML.

glogiotatidis:
about:home must be XML valid and since snippets are injected into
about:home, snippets must be XML valid. We already do XML validation of
the snippet template variables and with this change we do XML validation
of the templates too.


glogiotatidis:
@pmclanahan r?

pmclanahan:
Nice work :fireworks:

PR #133, opened 2015-11-16
Count clicks only in links within topSection.

glogiotatidis:
Do not handle `click`s outside topSection.

@schalkneethling r?

### DIFF
--- a/snippets/base/templates/base/includes/snippet_js.html
+++ b/snippets/base/templates/base/includes/snippet_js.html
@@ -444,33 +444,36 @@ Mozilla.UITour.setConfiguration = function(configName, configValue) {
         }
     }
 
-    // Listen for clicks on links that point to 'about:accounts' and
-    // convert them showFirefoxAccounts Events.
-    document.addEventListener('click', function(event) {
+    // Listen for clicks on links, send metrics and handle
+    // about:account custom links.
+    var topSection = document.getElementById('topSection');
+    topSection.addEventListener('click', function(event) {
         var target = event.target;
-        while (target && target.tagName && target.tagName.toLowerCase() !== 'a') {
+        while (target.tagName && target.tagName.toLowerCase() !== 'a') {
+            // Do not track clicks outside topSection.
+            if (target.id === 'topSection') {
+                return;
+            }
             target = target.parentNode;
         }
 
         // Count snippet clicks.
-        if (target && target.href && target.dataset.eventCounted !== 'true') {
+        if (target.dataset.eventCounted !== 'true') {
             target.dataset.eventCounted = 'true';
             // Fetch custom metric or default to 'click'
             var metric = target.dataset.metric || 'click';
-
             // If user is not opening a new tab, preventDefault action.
-            if (event.button === 0 && !(event.metaKey || event.ctrlKey)) {
+            if (target.href && (event.button === 0 && !(event.metaKey || event.ctrlKey))) {
                 event.preventDefault();
                 var callback = function() {
                     target.click();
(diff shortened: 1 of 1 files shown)


snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -444,33 +444,36 @@
...
-    document.addEventListener('click', function(event) {
+    // Listen for clicks on links, send metrics and handle
+    // about:account custom links.
+    var topSection = document.getElementById('topSection');
+    topSection.addEventListener('click', function(event) {
         var target = event.target;
-        while (target && target.tagName && target.tagName.toLowerCase() !== 'a') {
+        while (target.tagName && target.tagName.toLowerCase() !== 'a') {
+            // Do not track clicks outside topSection.
+            if (target.id === 'topSection') {
```
schalkneethling: Seeing that we are adding the `eventListener` to the element with `id`, `topSection` above, do we still need this check? There should not be more than one element on the page with an `id` of `topSection` correct?

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -444,33 +444,36 @@
...
             // If user is not opening a new tab, preventDefault action.
-            if (event.button === 0 && !(event.metaKey || event.ctrlKey)) {
+            if (target.href && (event.button === 0 && !(event.metaKey || event.ctrlKey))) {
                 event.preventDefault();
                 var callback = function() {
                     target.click();
                 }
             }
-
             sendMetric(metric, callback);
```
schalkneethling: I am wondering if we can write the above as follows:

```
if (target.href && (event.button === 0 && !(event.metaKey || event.ctrlKey))) {
    event.preventDefault();
    sendMetric(metric, function() {
        target.click();
    });           
}
```

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -444,33 +444,36 @@
...
         }
 
         // Handle about:accounts clicks.
-        if (target && target.href && target.href.indexOf('about:accounts') === 0) {
+        if (target.href && target.href.indexOf('about:accounts') === 0) {
```
schalkneethling: Won't the value of `href` in this case simply be `about:accounts`? If so, we probably do not need `indexOf`?

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Correct, only one element with id=topSection. What we do here is that we traverse the tree in reverse trying to find an `<a>`. We can start within `topSection` but because of `target=target.parentNode` get out of it and that's why we need this check, correct?

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: We still want to send the metric even if the user is opening a new tab.

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Nope, the link would have been altered by `modifyLinks` and it would be something like `about:accounts?snippet_name=33&sample_rate=0.1`

snippets/base/templates/base/includes/snippet_js.html
schalkneethling: Ah! Of course.

schalkneethling:
lgtm r+

glogiotatidis:
Thanks!

PR #135, opened 2015-11-24
fix bug 1227601 - replace fig with docker-compose

groovecoder:
How I was able to get everything set up using `docker-machine` and `docker-compose` instead of `boot2docker` and `fig`


glogiotatidis:
Replaced by #138 Thanks @groovecoder

PR #136, opened 2015-11-25
Add snippet ids in response.

glogiotatidis:
Since we're now sending snippets in a json encoded format it's harder to
identifiy which snippets are included. This commit add an html comment
with the ids of the snippets included in each response.


glogiotatidis:
Tests pass, this is minor change

PR #138, opened 2015-12-03
Rebase on sugardough.

glogiotatidis:
Rebase snippets to sugardough features:
 - Django LTS 1.8 
 - 12factorization

and opens the gates to a update-all-the-things and make snippets fun again


glogiotatidis:
@pmclanahan please r? It's only the basics needed to 12factorize snippets and upgrade to a recent django version + libs :sweat_smile:

glogiotatidis:
Yo @pmclanahan I know that you're probably too busy with the durable team. Let me know if you'd like me to request review from other mozillians.

Thanks!

pmclanahan:
Sorry Giorgos. I'll get to this today.

glogiotatidis:
> Sorry Giorgos. I'll get to this today.

:beers: :)

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends build-essential python python-dev python-pip \
+                                               libmysqlclient-dev libxslt1.1 libxml2 libxml2-dev libxslt1-dev && \
+    rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
 
+# Pin a known to work with peep pip version.
+RUN pip install pip==6.0.0
```
pmclanahan: With the newest peep you should be able to get rid of this. It support up through pip 8. I figured we'd eventually move this to pip 8 with pipstrap, but don't need anything more in this PR.

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
+
 # First copy requirements.txt and peep so we can take advantage of
 # docker caching.
-COPY requirements /tmp/requirements
```
pmclanahan: need to keep the copy of the `requirements` dir since `requirements.txt` below requires those files as well. The docker build isn't working for me due to that.

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
-RUN pip install -r /tmp/requirements/dev.txt
+COPY ./bin/peep.py /app/bin/peep.py
+COPY requirements.txt /app/requirements.txt
+RUN ./bin/peep.py install -r requirements.txt
 
-EXPOSE 8000
-ENV PYTHONUNBUFFERED 1
+COPY . /app
+RUN DEBUG=False SECRET_KEY=foo ALLOWED_HOSTS=localhost, DATABASE_URL=sqlite:/// ./manage.py collectstatic --noinput -c
```
pmclanahan: Need to include `SITE_URL` here or set a default in settings.

pmclanahan: also need to make sure the `static` dir exists for this to succeed.

glogiotatidis:
Yeah I wasn't trying this with Docker just with virtualenv, sorry for the trouble @pmclanahan. Fixed your comments.

pmclanahan:
Ah. Sorry. I saw changes to the docker stuff and thought I could run tests that way. The diff is too big to really review, so I'm trying to just put it through it's paces manually and see if it all works :)

pmclanahan:
This is fine I think. Tests pass in tox in Travis, and the changes are many but it seems to work. I say we merge it and move forward. Thanks!! Nice work :cake:

pmclanahan:
Could use a rebase before merge though.

glogiotatidis:
> Still doesn't work because we need SITE_URL set and the static folder needs to exist when using -c, though we probably just don't need -c. But if you're not yet using docker than we can fix this later.

Got it right by removing `-c`

glogiotatidis:
> This is fine I think. Tests pass in tox in Travis, and the changes are many but it seems to work. I say we merge it and move forward. Thanks!! Nice work :cake:

Thanks for your time pmac :) :mushroom:

PR #139, opened 2015-12-16
 [bug 1233001] Switch non release channels to MLS 

glogiotatidis:
GeoDude (geo.mozilla.org) will be decommissioned soon and the snippets service needs to switch to MLS for location info.

### DIFF
--- a/snippets/base/templates/base/includes/snippet_js.html
+++ b/snippets/base/templates/base/includes/snippet_js.html
@@ -1,5 +1,6 @@
 <script type="text/javascript">
 //<![CDATA[
+{% if client.channel == 'release' %}
 /*!
   * $script.js Async loader & dependency manager
   * https://github.com/ded/script.js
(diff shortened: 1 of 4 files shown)


glogiotatidis:
@pmclanahan r?

pmclanahan:
how soon? I think bedrock is still using it client side.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -355,19 +358,45 @@
...
-            try {
-                gSnippetsMap.set('geoCountry', geoip_country_code());
-                gSnippetsMap.set('geoLastUpdated', new Date());
-            } catch (e) {
-                // Most likely failed to load JS file. Continue on without us,
-                // we'll try again next time.
-            }
-        });
-    }
+    {% if client.channel == 'release' %}
```
pmclanahan: Will this have to change again on the next release? I assume "release" channel isn't going to stay on the old server for long.

pmclanahan:
Looks good to me, though I don't see any tests for the JS and I've not tested it myself. Is there a good way to test this locally?

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -355,19 +358,45 @@
...
+      }
+    {% else %}
+      // Download the user's country using the geolocation service.
+      // Please do not directly use this code or Snippets key.
+      // Contact MLS team for your own credentials.
+      // https://location.services.mozilla.com/contact
+      function downloadUserCountry() {
+          var GEO_URL = "{{ settings.GEO_URL }}";
+          var request = new XMLHttpRequest();
+          request.onreadystatechange = function() {
```
hannosch: maybe add ``request.timeout = 60000; // timeout after 60 seconds``

Networks are unreliable and I'd always add timeouts to any network request, so they don't stay around forever.

glogiotatidis:
> Looks good to me, though I don't see any tests for the JS and I've not tested it myself. Is there a good way to test this locally?

One way is to run the service and visit
`http://127.0.0.1:8000/4/sms/44.0a1/20151028030432/Linux_x86_64-gcc3/en-US/nightly/Linux%204.2.0-1-amd64%20%28GTK%203.18.2%29/default/default/`

to verify that nightly gets the new code

and 

`http://127.0.0.1:8000/4/sms/44.0a1/20151028030432/Linux_x86_64-gcc3/en-US/release/Linux%204.2.0-1-amd64%20%28GTK%203.18.2%29/default/default/`

to verify that release gets the old code.

If you have a firefox profile setup to read snippets from your local instance you can use the DevTools console to verify that the request goes through.

For the record, I've tested this on nightly (current) and on release (Fx 15 to current)

We don't have any JS tests yet :|

glogiotatidis:
> how soon? I think bedrock is still using it client side.

I think early 2016. Ping atoll for info on the current situation and @hannosch to get an API key for MLS

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -355,19 +358,45 @@
...
+                      gSnippetsMap.set('geoCountry', country_data.country_code);
+                      gSnippetsMap.set('geoLastUpdated', new Date());
+                  } catch (e) {
+                      // Most likely failed to load Data file. Continue on without us,
+                      // we'll try again next time.
+                  }
+              }
+          };
+          request.open('GET', GEO_URL, true);
+          request.send();
```
hannosch: Over at https://dxr.mozilla.org/mozilla-central/source/toolkit/components/search/nsSearchService.js#695 I've seen:
```
request.setRequestHeader("Content-Type", "application/json");
request.responseType = "json";
```
which might result in automatic parsing of the response, so ``response.country_code`` is available and you can avoid the extra JSON.parse call.

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Nice catch, added.

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Interesting. In this case we need to make changes to the server to allow setting the header. Here's the error I'm getting:
```Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://location.services.mozilla.com/v1/country?key=fff72d56-b040-. (Reason: missing token 'content-type' in CORS header 'Access-Control-Allow-Headers' from CORS preflight channel).```

Setting ```Access-Control-Request-Headers:"content-type"``` in the response headers should do it.

If that's easy to do let's go for it, otherwise we can keep JSON.parse.

hannosch: @glogiotatidis ok, then ignore my suggestion for now. I'll look into the additional header in more detail next year. We won't do any new non-emergency deployments to the service side code this year. Since your version works now, let's go with that so you can switch this on whenever you want to :)

glogiotatidis:
Updated with review comments.

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: ack!

hannosch:
r+ looks good to me

snippets/base/templates/base/includes/snippet_js.html
glogiotatidis: Discussed in IRC but answering also here for the record: 

This is to limit the impact of this change to non-release users and not to a specific Fx version. It will give us the opportunity to test and verify on beta/aurora/nightly and make sure that MLS can handle the load.

glogiotatidis:
Thanks both!

PR #141, opened 2016-01-04
Add campaign based opt outs.

ScottDowne:
Playing around a bit.

I notice there was a campaign prop in the snippet service, so I was wondering if we could use that to get better opt out surface area. Code was fairly simple so I provided that as an example.

Example, a snippet with no campaign is opt'ed out, so it uses the id. However, if a campaign is present, example "2015-eoy", it opts me out of all snippets under that campaign name.

I'm not super familiar with the historical usage of the campaign name, and I see a potential issue in a campaign name being re used by accident. As it's not an ID and nothing is forcing it to be unique, that I know of. Looks like currently it's used for metrics.

### DIFF
--- a/snippets/base/templates/base/includes/snippet_js.html
+++ b/snippets/base/templates/base/includes/snippet_js.html
@@ -248,7 +248,10 @@ Mozilla.UITour.setConfiguration = function(configName, configValue) {
         var blockList = getBlockList();
         snippets = snippets.filter(
             function (snippet) {
-                if (blockList.indexOf(snippet.id) === -1) {
+                var snippet_id = parseInt(snippet.id, 10);
+                var campaign = snippet.campaign;
+                var snippetHandle = campaign || snippet_id;
+                if (blockList.indexOf(snippetHandle) === -1) {
                     return true;
                 }
                 return false;
@@ -425,10 +428,15 @@ Mozilla.UITour.setConfiguration = function(configName, configValue) {
 
     // Add links to buttons that block snippets.
     function addSnippetBlockLinks(elements) {
-        var snippet_id = ABOUTHOME_SHOWN_SNIPPET.id;
+        var snippet_id = parseInt(ABOUTHOME_SHOWN_SNIPPET.id, 10);
+        var campaign = ABOUTHOME_SHOWN_SNIPPET.campaign;
+        var snippetHandle = campaign || snippet_id;
+        if (!snippetHandle && snippetHandle !== 0) {
+            return;
+        }
         var blockSnippet = function (event) {
             event.preventDefault();
-            addToBlockList(snippet_id);
+            addToBlockList(snippetHandle);
             sendMetric('snippet-blocked', function() { window.location.reload(); })
         };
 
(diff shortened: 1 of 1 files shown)


bensternthal:
Scot - Just an FYI @glogiotatidis  who owns snippets is on PTO until the 11th. Assuming this is OK to wait he will review when he gets back.

ScottDowne:
Yup, no rush :)

glogiotatidis:
@ScottDowne this is great stuff. :gem:  Please hang tight while I'm fighting with tasks that surfaced while on PTO and then I'll be mainly focused on snippets for a while.

ScottDowne:
No prob. I'm in a similar position.

ScottDowne:
I am a bit worried this increases the potential for problematic campaign name conflicts. Not a lot, just a bit.

Some brainstorming for future ideas.

I think ideally the campaign names would be chosen from a list, like a drop down (or even cooler a datalist), so the act of picking one is very explicit. Adding a new one would also be possible. At this point it doesn't need to even be the the campaign variable, can be something new. Maybe works more like a tag?

This would give us the added bonus of exposing previous values when someone else ships a new snippet, so if I'm adding a similar snippet to a previous one, I would be able to use an existing value, if I want, and not by accident or in an unexpected way.

I wasn't able to dig into enough of the codebase yet to figure that out, and it also seemed like more work, for less impact, compared to what I was already able to do with little work. It also seemed like an enhancement and not core functionality, so maybe if we want something like this we can add it later.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -248,7 +248,10 @@
         var blockList = getBlockList();
         snippets = snippets.filter(
             function (snippet) {
-                if (blockList.indexOf(snippet.id) === -1) {
+                var snippet_id = parseInt(snippet.id, 10);
+                var campaign = snippet.campaign;
+                var snippetHandle = campaign || snippet_id;
+                if (blockList.indexOf(snippetHandle) === -1) {
```
glogiotatidis: We need to test both against `campaign` and `snippet_id` to be backwards compatible.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -425,10 +428,15 @@
 
     // Add links to buttons that block snippets.
     function addSnippetBlockLinks(elements) {
-        var snippet_id = ABOUTHOME_SHOWN_SNIPPET.id;
+        var snippet_id = parseInt(ABOUTHOME_SHOWN_SNIPPET.id, 10);
+        var campaign = ABOUTHOME_SHOWN_SNIPPET.campaign;
+        var snippetHandle = campaign || snippet_id;
+        if (!snippetHandle && snippetHandle !== 0) {
```
glogiotatidis: Why is this needed?

glogiotatidis:
This needs a rebase

glogiotatidis:
Also please update documentation https://github.com/mozilla/snippets-service/blob/master/docs/developing.rst#snippet-block-list

glogiotatidis:
It also needs a comment (help_text) in admin panel, under campaign field to remind that this will be used for the block list

glogiotatidis:
I want to make sure that the snippet editors understand the modifications in this pull request, so I'll communicate this further before merging.

If you have the time please rebase and fix review comments but no need to prioritize this one.

Thanks!

glogiotatidis:
Closing in favor of #222. Thanks for your work @ScottDowne

PR #142, opened 2016-01-06
Manually update snippet after opt out to avoid refreshing the page.

ScottDowne:


ScottDowne:
Main things I'm doing here:

1. Removed usage of localStorage. I don't see how this ever worked in about:home, but it can be added back if needs be.
2. After an opt out, instead of refreshing the page to update the displayed snippet, it now removes the snippet and displays a new one manually.

ScottDowne:
This fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1236090

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -117,66 +117,61 @@
...
-    } else {
-        // localStorage isn't available, use gSnippetsMap (backed by IndexedDB).
-        gSnippetsMap = window.gSnippetsMap;
-    }
+    var gSnippetsMap = window.gSnippetsMap;
+
+    // By the time show_snippet event reaches #snippets the snippet
+    // will have finished initializing and altering the DOM. It's the
+    // time to modifyLinks() and addSnippetBlockLinks().
+    var snippets = document.getElementById('snippets');
```
ScottDowne: This event used to be created in the middle of injecting the snippet.

Now that the inject snippet block is in a function, and this only needs to happen once, and not whenever a snippet is injected, I moved it up a bit.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -117,66 +117,61 @@
...
+    snippets.addEventListener('show_snippet', function(event) {
```
ScottDowne: w

This event used to be created in the middle of injecting the snippet.

Now that the inject snippet block is in a function, and this only needs to happen once, and not whenever a snippet is injected, I moved it up a bit.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -117,66 +117,61 @@
...
+        var topSection = document.getElementById('topSection');
+        // Add sample rate and snippet ID to currently displayed links.
+        var parameters = ('sample_rate=' + SNIPPET_METRICS_SAMPLE_RATE + '&snippet_name=' +
+                          ABOUTHOME_SHOWN_SNIPPET.id);
+        modifyLinks(topSection.querySelectorAll('a'), parameters);
+        addSnippetBlockLinks(topSection.querySelectorAll('.block-snippet-button'));
+        sendImpression();
+    });
 
+    var snippetContainer;
```
ScottDowne: Store this element outside the `injectSnippet` scope so `injectSnippet` and `removeSnippet` can share its reference.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -117,66 +117,61 @@
...
+      } else {
+          showDefaultSnippets();
+      }
+    }
 
-        // By the time show_snippet event reaches #snippets the snippet
-        // will have finished initializing and altering the DOM. It's the
-        // time to modifyLinks() and addSnippetBlockLinks().
-        var snippets = document.getElementById('snippets');
-        snippets.addEventListener('show_snippet', function(event) {
```
ScottDowne: This block has been moved, not deleted.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -428,8 +423,16 @@
         var snippet_id = ABOUTHOME_SHOWN_SNIPPET.id;
         var blockSnippet = function (event) {
             event.preventDefault();
+            // We want to stop other click events.
+            // This element is being removed, so another
+            // click might cause parentNode reference errors.
+            // See trackLinkClicks()
+            event.stopPropagation();
```
ScottDowne: `trackLinkClicks` function is called whenever `topSection` is clicked, but its `useCapture` is `true` which was causing issues when firing the event after the snippet was removed.

I covered it here, but also just in case `trackLinkClicks` was also updated to handle the case where the element was removed from the DOM.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -465,11 +456,11 @@
     // Listen for clicks on links, send metrics and handle
     // about:account custom links.
     var topSection = document.getElementById('topSection');
-    topSection.addEventListener('click', function(event) {
+    function trackLinkClicks(event) {
         var target = event.target;
         while (target.tagName && target.tagName.toLowerCase() !== 'a') {
             // Do not track clicks outside topSection.
-            if (target.id === 'topSection') {
+            if (target.id === 'topSection' || !target.parentNode) {
```
ScottDowne: If we don't have a `parentNode`, we're no longer in the DOM and thus can return.

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -465,11 +456,11 @@
     // Listen for clicks on links, send metrics and handle
     // about:account custom links.
     var topSection = document.getElementById('topSection');
-    topSection.addEventListener('click', function(event) {
+    function trackLinkClicks(event) {
```
ScottDowne: I gave this a name simply to make it easier to reference in a comment earlier.

glogiotatidis:
Thanks for working on this Scott.

Avoiding the refresh will increase (but not eliminate) the IndexedDB errors and help with snippet blocking. 

Unfortunately many snippet templates are altering the DOM outside the `#snippets` container, by changing logo, body background colors etc. Your approach here will change what's within the #snippets container but will leave everything else unaffected. 

The best option to start fresh is to reload. Alternatively we can start requiring snippets that allow blocking to provide  a `restore()` function that will revert will changes made, but this sounds error prone.

Thoughts?

ScottDowne:
Yeah. I agree a restore function is not great. Also yeah a 500 Ms timeout
would work.

The main reason I went with the content change and not the relead is
because of where the indexdb code lived it would be more code changed to
add a callback for dB writes than to update the content. It's totally
doable though and given the reasons and options I thinks it's our best bet.

ScottDowne:
If there is a rush on this, the 500 ms timeout would probably be worth it.

ScottDowne:
I played around with this a bit.

I think there is some value in a restore function, but still not completely sold on it. I think if it's provided as an event that's the mostly the inverse of the existing 'show_snippet' event, as a 'snippet_removed' or something, it might be a bit better. A bit less prone to error?

I'm also playing around with the idea of callback based usage of indexeddb. The tricky bit here is currently it uses gSnippetsMap for blocked snippets, which has two issues.

1. gSnippetsMap uses cache and fundamentally works sync because of the cache. This is a nice feature for what it was being used for, so I'm hesitant to change it.
2. gSnippetsMap is shipped in the browser itself, so if we make changes there, we'll need to worry about older versions.

Feels like gSnippetsMap is serving another purpose.

I'm leaning towards are specific indexeddb just for blocked snippets and it'll live right where we need it. It's just a bit more code, and I'm kinda new to indexddb, so I also need to spend a bit of time to learn it, which is fine if it's the direction we go.

I see the 500ms timeout as viable only if we're in a rush. It would be really easy to update my patch to do that, so that's a plus.

glogiotatidis:
> I'm leaning towards are specific indexeddb just for blocked snippets and it'll live right where we need it. It's just a bit more code, and I'm kinda new to indexddb, so I also need to spend a bit of time to learn it, which is fine if it's the direction we go.

We could open a new connection to the current idb since we already have data there.

ScottDowne:
I tried. The reference is behind an API and the API doesn't support
callbacks. Requiring a patch in the browser bits.

glogiotatidis:
> I tried. The reference is behind an API and the API doesn't support callbacks. Requiring a patch in the browser bits.

Please elaborate, i don't really understand what you mean

ScottDowne:
While writing up a response, I figured out what you initially meant by "open a new connection"

I thought you meant use the existing db reference here: http://lxr.mozilla.org/mozilla-central/source/browser/base/content/abouthome/aboutHome.js#175

Which is behind the gSnippetsMap's set/get functions.

But, I think what you were saying is do this part again: http://lxr.mozilla.org/mozilla-central/source/browser/base/content/abouthome/aboutHome.js#125

I didn't know that was possible, but actually makes some sense! Is that what you meant?

glogiotatidis:
> But, I think what you were saying is do this part again: http://lxr.mozilla.org/mozilla-central/source/browser/base/content/abouthome/aboutHome.js#125
> 
> I didn't know that was possible, but actually makes some sense! Is that what you meant?

Yes!

ScottDowne:
@glogiotatidis I have a patch that seems to be doing what we need.

I made it a new pull request because I had rebase issues and this ended up being easier.

https://github.com/mozilla/snippets-service/pull/169

Closing this one.

PR #144, opened 2016-02-08
[bug 1238092] Prevent search box from stealing keypresses.

glogiotatidis:
See also https://github.com/mozilla/snippets/pull/77

### DIFF
--- a/snippets/base/templates/base/includes/snippet_js.html
+++ b/snippets/base/templates/base/includes/snippet_js.html
@@ -157,6 +157,7 @@ Mozilla.UITour.setConfiguration = function(configName, configValue) {
                               ABOUTHOME_SHOWN_SNIPPET.id);
             modifyLinks(topSection.querySelectorAll('a'), parameters);
             addSnippetBlockLinks(topSection.querySelectorAll('.block-snippet-button'));
+            keypressIsMine(snippetContainer.querySelector('.snippet'));
             sendImpression();
         });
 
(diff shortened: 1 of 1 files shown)


glogiotatidis:
@schalkneethling r? :)

snippets/base/templates/base/includes/snippet_js.html
```diff
@@ -411,6 +412,21 @@
...
 
+    // Bug 1238092. Got fixed in Fx45. We can drop this when Fx44
+    // usage drops significantly.
+    function keypressIsMine(snippet) {
+        function onKeyPress(e) {
+            e.stopPropagation();
+        }
+        function attachHandler(matches) {
+            for (var i = 0, len = matches.length; i < len; i++) {
+                matches[i].addEventListener('keypress', onKeyPress);
```
schalkneethling: probably not such a big deal but, if it is possible that there might be a few input and select elements, it can end up attaching a lot of event handlers. If this is not the case and the number is low 3-4 perhaps, then not to worry.

If it can get higher, then we should perhaps rather attach an event handler to the parent element and, if the `event.target` of the `keypress` event was an `input` or `select` we call `onKeyPress`, else we just let it pass through.

Something like:

```
parent.addEventListener('keypress', function(event) {
    var tagName = event.target.tagName;

    if (tagName === "INPUT" || tagName === 'SELECT') {
        onKeyPress();
    }
});
```

glogiotatidis: I don't think that we'll ever go above 3-4 input elements but maybe we can attach it to `.snippet`?

schalkneethling: That will work.

PR #146, opened 2016-02-08
Ignore 404 in FetchSnippetsMiddleware.

glogiotatidis:
FetchSnippetsMiddleware tries to resolve request.path which can raise a
404 if the path doesn't exist. The exception should be ignored to allow
the rest of the middlewares to run and handle the 404. This will allo
Django to append slash to URLs and fix the /admin VS /admin/ problem.


glogiotatidis:
@akatsoulas r?

akatsoulas:
looking good. r+ :cake:

glogiotatidis:
Thnx!

PR #148, opened 2016-02-08
Deploy to Deis

glogiotatidis:
Travis configuration and scripts plus code changes needed to run on Deis.


snippets/settings.py
```diff
@@ -185,3 +184,6 @@
 }
 
 GEO_URL = 'https://location.services.mozilla.com/v1/country?key=fff72d56-b040-4205-9a11-82feda9d83a3'  # noqa
+
+PROD_DETAILS_STORAGE = config('PROD_DETAILS_STORAGE',
+                              default='product_details.storage.PDFileStorage')
```
glogiotatidis: The plan is to enable the db backend and at cron job to update it regularly.

glogiotatidis:
@pmclanahan Please r?

snippets/settings.py
```diff
@@ -44,6 +43,7 @@
     'smuggler',
     'django_filters',
     'django_ace',
+    'product_details',
```
pmclanahan: I see django-redis installed but don't see any updates to settings for that. Is that missing or all just handled via the env?

pmclanahan: Yeah, I guess you can just set `CACHE_URL` to a `redis://` URL and it'll work. Nice :)

pmclanahan:
LGTM! r+ :snowman:

glogiotatidis:
Thanks @pmclanahan !

PR #149, opened 2016-02-08
[bug 1241124] Prevent search box from stealing keypresses.

glogiotatidis:
For legacy snippets.

PR #150, opened 2016-02-08
[bug 1241124] Prevent search box from stealing keypresses.

glogiotatidis:
Same as #149 but for new master


glogiotatidis:
Per review in #144 I'm merging this.

PR #152, opened 2016-02-08
Add option to disable admin

glogiotatidis:
This option will be needed in the instances pointing to the read-only db
replica to prevent users from logging in and trying to make changes.

Also fix the admin title 'cause why not?


glogiotatidis:
@akatsoulas r?

akatsoulas:
Tested this locally and work OK. r+ :confetti_ball:

glogiotatidis:
Thanks!

PR #157, opened 2016-02-11
Deploy also to eu-west

glogiotatidis:
@pmclanahan r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,16 @@ language: python
 env:
   global:
     # NewRelic
-    - NEWRELIC_STAGE_APPLICATION=snippets-stage-us-west
-    - NEWRELIC_PROD_APPLICATION=snippets-prod-us-west
+    - NEWRELIC_US_WEST_STAGE_APPLICATION=snippets-stage-us-west
+    - NEWRELIC_US_WEST_PROD_APPLICATION=snippets-prod-us-west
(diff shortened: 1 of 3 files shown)


pmclanahan:
Looks good. A couple of notes:

1. I like consolidating so that more work is done in the script and less in the config, but that's a personal preference. i.e. I'd have just passed in "eu-west" or "us-west" and used that construct the deis endpoint and new relic app names. And I do like my loop that does USW and EUW in a single script call, but your way is more flexible.
2. I like giving the positional argument variables (e.g. `$1`) descriptive names in the script so that I don't have to keep going back to the travis config to remember which is which. But again that's just a preference.

These are not required changes obviously. It's great just like it is. r+

glogiotatidis:
> I like consolidating so that more work is done in the script and less in the config, but that's a personal preference. i.e. I'd have just passed in "eu-west" or "us-west" and used that construct the deis endpoint and new relic app names. And I do like my loop that does USW and EUW in a single script call, but your way is more flexible.
I see your point but I prefer to have this in the config.

> I like giving the positional argument variables (e.g. $1) descriptive names in the script so that I don't have to keep going back to the travis config to remember which is which. But again that's just a preference.
Good point, changed that.


Thanks for the good suggestions @pmclanahan  :) :v:

PR #158, opened 2016-02-12
Add env variable to control number of wsgi workers.

glogiotatidis:
@jgmize Please r?


jgmize:
After I applied the fix in #159, I tried running this branch with
```sh
docker-compose run web bin/run-prod.sh
```
and I noticed the following in the output:
```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
bin/run-prod.sh: 5: bin/run-prod.sh: cannot create static/revision.txt: Directory nonexistent
```

Both of those are outside of the scope of this PR, and may already be known issues, but I wanted to point them out here just in case.

I tested overriding WSGI_NUM_WORKERS and WSGI_LOG_LEVEL in the docker-compose.yml and those both worked as expected. r+, nice work.

PR #160, opened 2016-02-15
Support uploading bundles to S3.

glogiotatidis:
- [x] need to support different cache headers per type of file


glogiotatidis:
Hey @jgmize if you could review this before today's load test session it would be great. Thanks!

snippets/base/migrations/0002_auto_20160215_1300.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('base', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='uploadedfile',
+            name='file',
+            field=models.FileField(upload_to=snippets.base.models._generate_filename),
```
jgmize: This migration doesn't appear to actually do anything at the db level, it just removes the `storage` kwarg present in the `0001_initial` migration. It seems safe, but we should validate that it works with `settings.DEFAULT = 'storages.backends.overwrite.OverwriteStorage'` in the legacy staging environment-- not a blocker for merge, just noting here.

snippets/base/storage.py
```diff
@@ -7,3 +15,39 @@ def get_available_name(self, name):
...
+class S3Storage(S3BotoStorage):
+    cache_control_headers = getattr(settings, 'AWS_CACHE_CONTROL_HEADERS', {})
+
+    def _save(self, name, content):
+        cleaned_name = self._clean_name(name)
+        name = self._normalize_name(cleaned_name)
+        headers = self.headers.copy()
+        for filename_start, value in self.cache_control_headers.iteritems():
+            if name.startswith(filename_start):
+                headers['Cache-Control'] = value
```
jgmize: The Cache-Control header appears to be the only difference from the current https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py -- it seems like a reasonable change, but is there (or should there be) a PR against the main repo to incorporate this upstream?

snippets/base/migrations/0002_auto_20160215_1300.py
jgmize: While looking through the open PRs against django-storages-redux, I noticed https://github.com/jschneier/django-storages/pull/111 which looks like a promising upgrade.

PR #162, opened 2016-02-16
Add remove image link in admin.

glogiotatidis:
Allow snippet editors to remove images added in image fields.

PR #163, opened 2016-02-16
Add remove image link in admin.

glogiotatidis:
Allow snippet editors to remove images added in image fields.

### DIFF
--- a/snippets/base/static/js/templateDataWidget.js
+++ b/snippets/base/static/js/templateDataWidget.js
@@ -109,6 +109,13 @@
                 self.onImageFieldChange(this);
             });
 
+            this.$container.on('click', '.image-input-remove', function(event) {
+                event.preventDefault();
+                var imageInput = $(event.target).siblings('input')[0];
+                $(imageInput).val('');
+                self.onImageFieldChange(imageInput);
+            });
+
             this.$container.on('input', 'textarea', function() {
                 self.triggerDataChange();
             });
@@ -161,7 +168,10 @@
          */
         onImageFieldChange: function(input) {
             var self = this;
+            var preview = $(input).siblings('img')[0];
             if (input.files.length < 1) {
+                preview.src = '';
+                self.triggerDataChange();
                 return;
             }
 
@@ -180,7 +190,6 @@
             }
 
             // Load file.
-            var preview = $(input).siblings('img')[0];
             var reader = new FileReader();
             reader.onload = function(e) {
                 preview.src = e.target.result;
--- a/snippets/base/static/templates/snippetDataWidget.html
+++ b/snippets/base/static/templates/snippetDataWidget.html
@@ -12,6 +12,7 @@
         {% elif variable.type == types.image %}
           <img src="{{ originalData[variable.name] }}">
           <input type="file" class="image-input">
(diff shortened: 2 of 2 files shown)


glogiotatidis:
Hey @ScottDowne, do you have time to review this one?

glogiotatidis:
Ping @ScottDowne r?

ScottDowne:
Hey, yeah, I do. I saw this go by while on vacation then forgot.

First question while I get started is what's the difference between this pr and #162?

snippets/base/static/js/templateDataWidget.js
```diff
@@ -161,7 +168,10 @@
          */
         onImageFieldChange: function(input) {
             var self = this;
+            var preview = $(input).siblings('img')[0];
```
ScottDowne: This var is created already later in this function.

snippets/base/static/templates/snippetDataWidget.html
```diff
@@ -12,6 +12,7 @@
         {% elif variable.type == types.image %}
           <img src="{{ originalData[variable.name] }}">
           <input type="file" class="image-input">
+          <a href="#" class="image-input-remove">Remove image</a>
```
ScottDowne: Just a nit, probably doesn't really matter:

If easy to add a 5px left margin/padding to make it line up with the browse button?

ScottDowne:
@glogiotatidis that's all I could find. Looks good.

ScottDowne:
I also noticed, but was unrelated to this patch is a console error while loading the snippet editor:

TypeError: data is undefined: templateDataWidget.js:88:25

Also displaying hiding an image element when the image src is empty is easy, instead of displaying a broken link: `img[src=""] {
    display: none;
}` or something.

snippets/base/static/js/templateDataWidget.js
glogiotatidis: I moved it from line 183 up here. Do I miss something?

ScottDowne: That should do it!

snippets/base/static/templates/snippetDataWidget.html
glogiotatidis: Good point, added in #162

glogiotatidis:
> I also noticed, but was unrelated to this patch is a console error while loading the snippet editor:
> 
> TypeError: data is undefined: templateDataWidget.js:88:25
> 
> Might be just for me, might not be reproducible for you.
> 
> Also hiding an image element when the image src is empty, instead of displaying a broken link, is easy: img[src=""] { display: none; } or something.

Fixed in #162 as well.

glogiotatidis:
> First question while I get started is what's the difference between this pr and #162?

We currently have two diverged branches. On production atm is `legacy` and we're moving soon production to `master` which features a bunch of library upgrades and fixes. I try to bring needed functionality to both branches thus the two identical PRs against the two different branches. 

Beautification comments and other fixes I made in #162 to bring to the new platform. I'll keep this one just for the removeImage functionality.

glogiotatidis:
Thanks for your time @ScottDowne. Do you mind spending some time reviewing #162 that resolves the rest of your review comments?

PR #166, opened 2016-02-24
Add clear_cache Django management command.

glogiotatidis:
Sometimes you just need to clear the cache.


glogiotatidis:
@pmclanahan r?

pmclanahan:
Why not :)

PR #167, opened 2016-02-24
Make BUNDLES and FILES root configurable.

glogiotatidis:
We can set `us-west` instances to save to `bundles/*` directory in their own `us-west` S3 bucket and the `eu-west` instances to save to `eu-west/bundles/*` directory in the own `eu-west` S3 buckets.

Then we can have one CloudFront CDN with two origins, the `eu-west` and the `us-west` S3 buckets and create a `Behavior` to fetch all `eu-west/*` matching files from `eu-west` and the rest from `us-west`. 

How does that sound @jgmize?

### DIFF
--- a/snippets/base/models.py
+++ b/snippets/base/models.py
@@ -8,7 +8,7 @@ import xml.sax
 from StringIO import StringIO
 from collections import namedtuple
 from datetime import datetime
-from urlparse import urljoin
+from urlparse import urljoin, urlparse
 from xml.sax import ContentHandler
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


jgmize:
Nice work, @glogiotatidis, r+. This seems like a reasonable approach, and the code changes look good to merge and test once we have the cloudfront config in place. Once we've confirmed that's all working as expected I'd love to see some docs to help explain this setup (ideally including an overview diagram depicting each deployment reading from its own db, writing to its own s3 bucket, and the whole thing fronted by the cdn) and how to handle some basic failover scenarios.

glogiotatidis:
Yap will be documented in detail in the migration plan and the snippets mana page 

Thanks @jgmize

PR #168, opened 2016-02-24
Make snippet bundle flag a setting, remove waffle.

glogiotatidis:
One less middleware, one less db hit, one less dependency. Good times!

### DIFF
--- a/snippets/settings.py
+++ b/snippets/settings.py
@@ -39,7 +39,6 @@ INSTALLED_APPS = [
 
     # Third party apps
     'django_jinja',
-    'waffle',
     'smuggler',
     'django_filters',
     'django_ace',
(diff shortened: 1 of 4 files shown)


glogiotatidis:
@jgmize r?

snippets/settings.py
```diff
@@ -171,6 +165,8 @@
     'default': config('CACHE_URL', default='locmem://', cast=django_cache_url.parse),
 }
 
+SERVE_SNIPPET_BUNDLES = config('SERVE_SNIPPET_BUNDLES', default=False, cast=bool)
```
jgmize: Wouldn't it make more sense to default this to either `True` or `not DEBUG`?

jgmize:
r+wc, nice work. If you decide to leave the default as-is we just need to make sure that staging and production have the new `SERVE_SNIPPET_BUNDLES` environment variable.

snippets/settings.py
glogiotatidis: I like `not DEBUG`. /me changes default

glogiotatidis:
> r+wc, nice work. If you decide to leave the default as-is we just need to make sure that staging and production have the new SERVE_SNIPPET_BUNDLES environment variable.

ACK

PR #169, opened 2016-02-29
iDB callback after snippet block

ScottDowne:


ScottDowne:
I need to figure out how to rebase on master, but if I update to master right now my local dev breaks so I havn't taken the time to figure that out yet.

ScottDowne:
@glogiotatidis hm

Trying to rebase this, but first updating to master, and hitting an issue with running `docker-compose up`

I get errors that `SITE_URL` is not defined. I added a `    - SITE_URL=http://localhost:8000` line under environment in docker-compose.yml, which seemed to fix it. Is this something that was simply missed or did I miss something?

ScottDowne:
@glogiotatidis Once I have it setup, if I go to localhost:8000/admin/ I have no idea what the default credentials are to log in as an admin.

ScottDowne:
@glogiotatidis I created an admin account with this command `docker-compose run web ./manage.py createsuperuser` which was easy once I found it, however, wasn't super obvious and I needed to dig to look it up.

ScottDowne:
@glogiotatidis ok, rebased this and cleaned it up a bunch and made the diff a bit easier to read.

glogiotatidis:
This was good work but we didn't proceed because we implemented a workaround. Now we're gradually switching to AS where we can use promises. 

Thank for you time @ScottDowne 🥇

PR #173, opened 2016-03-02
stopPropagation while block snippet button gets clicked.

glogiotatidis:
In some snippet templates which can become globally clickable (i.e.
snippet as a button) the block button in included in a global <a>
element. In some cases and FF versoisn when clicking the block button
the event propages all the way to the <a> and gets the user in another
page. stopPropagation fixes this.


glogiotatidis:
Copy of #172  for legacy codebase

glogiotatidis:
@schalkneethling r?

schalkneethling:
r+

We should create a function like this, to use in these situations:

```
function thouShallNotPass(event) {
   event.preventDefault();
   event.stopPropagation();
}
```

PR #175, opened 2016-03-09
Add clock job to update product details.

glogiotatidis:
Copy / pasta from lumbergh. @pmclanahan please r? :)


scripts/cron.py
```diff
@@ -0,0 +1,69 @@
...
+        try:
+            self.callback()
+        except Exception as e:
+            self.log('CRASHED: {}'.format(e))
+            raise
+        else:
+            self.log('finished successfully')
+
+    def log(self, message):
+        msg = '[{}] Clock job {}@{}: {}'.format(
```
pmclanahan: Looks like due to the way deis handles logs that the date here will be redundant as it's added automatically by the logging service. So in papertrail this will be duplicate info. Not sure if you care or not, but I think it can be removed if you'd like, both here and in lumbergh. But not a blocker for merge IMHO.

pmclanahan:
LGTM! r+

But looks like there are merge conflicts.

scripts/cron.py
glogiotatidis: Indeed but it's convenient when you debug locally. I'll find a better way to do this in the future.

glogiotatidis:
Thanks pmac!

PR #185, opened 2016-04-04
Fix /robots.txt

glogiotatidis:
* Add default value for ENGAGE_ROBOTS
 * s/mimetype/content_type/


glogiotatidis:
@jgmize please r? :D

PR #184, opened 2016-04-01
Refactor countries and locales.

glogiotatidis:

PR #193, opened 2016-04-20
Remove legacy locale and country code.

glogiotatidis:
This is removing already inactive code from #184


glogiotatidis:
Tests pass, merging.

PR #194, opened 2016-04-20
Remove django_smuggler.

glogiotatidis:
We don't longer need this. Devs have direct access to the database if and when needed.


glogiotatidis:
Tests pass, merging.

PR #195, opened 2016-04-21
 [fix bug 933848] Allow snippet filtering based on version. 

glogiotatidis:


glogiotatidis:
@akatsoulas r?

glogiotatidis:
Hey @ScottDowne what do you think about that r?

snippets/base/forms.py
```diff
@@ -151,6 +167,22 @@ class Meta:
...
+        version_lower_bound = cleaned_data['client_option_version_lower_bound']
+
+        if (version_upper_bound == 'older_than_current_release' and
+                version_lower_bound == 'current_release'):
+            raise forms.ValidationError(
+                "It doesn't make sense to combine those two Firefox version filters")
+
+        if 'any' not in [version_lower_bound, version_upper_bound]:
+            if Version(version_upper_bound) < Version(version_lower_bound):
+                raise forms.ValidationError('Firefox foo')
```
akatsoulas: Maybe something more descriptive?

glogiotatidis: erm yes obviously

snippets/base/forms.py
```diff
@@ -119,6 +121,14 @@ class BaseSnippetAdminForm(forms.ModelForm):
...
     template = forms.ModelChoiceField(queryset=SnippetTemplate.objects.exclude(hidden=True),
                                       widget=TemplateSelect)
+    client_option_version_lower_bound = forms.ChoiceField(
+        label='Firefox Version at least',
+        choices=[('any', 'No limit'),
+                 ('current_release', 'Current release')])
+    client_option_version_upper_bound = forms.ChoiceField(
+        label='Firefox Version at most',
+        choices=[('any', 'No limit'),
+                 ('older_than_current_release', 'Older than current release')])
```
akatsoulas: Why don't you use the same name as in the `lower_bound` option, eg `Current release`? The way it is right now can lead to confusion since it makes sense to assume that `Older than current release < Current release`.

akatsoulas: Well I am a bit confused here, if `current release == Older than current release`, I would suggest to change the name. If not please ignore :)

snippets/base/forms.py
```diff
@@ -151,6 +167,23 @@ class Meta:
             'data': TemplateDataWidget('template'),
         }
 
+    def clean(self):
+        cleaned_data = super(SnippetAdminForm, self).clean()
+        version_upper_bound = cleaned_data['client_option_version_upper_bound']
+        version_lower_bound = cleaned_data['client_option_version_lower_bound']
+
+        if (version_upper_bound == 'older_than_current_release' and
```
akatsoulas: You could move this check inside the `if 'any' clause` if you would like.

snippets/base/forms.py
glogiotatidis: Well `current release != older than current release`. Keep in mind that boundaries are inclusive, e.g. if you set the boundary to 38 this includes version 38.

Similarly `older than current release` includes everything older than current release and not current release.

snippets/base/views.py
```diff
@@ -114,11 +116,14 @@ def fetch_render_snippets(request, **kwargs):
                          .select_related('template')
                          .filter_by_available())
 
+    current_firefox_version = (
+        version_list(product_details.firefox_history_major_releases)[0].split('.', 1)[0])
```
akatsoulas: nit: You could also use the `LooseVersion/StrictVersion` from `distutils.version` if you feel that helps the readability:

```python
v = max(map(StrictVersion, product_details.firefox_history_major_releases.keys()))
v.version
```

glogiotatidis: I'll just stick with `split` instead of `map`ing and `max`ing !

akatsoulas:
r+(wc?) :beers:

glogiotatidis:
Thanks for the thorough review @akatsoulas :)

PR #205, opened 2016-07-19
Upgrade Django to version 1.8.14.

akatsoulas:


glogiotatidis:
Thanks @akatsoulas. We actually did the same thing with a couple of minutes difference at #206 .

PR #207, opened 2016-07-25
Make ESR channel follow Release channel.

glogiotatidis:
By reading HTTP logs we figured that ESR versions report ESR channel instead of release. With this patch we're treating ESRs as Release.

@pmac r?


pmac:
r+ 👍

glogiotatidis:
Merci!

PR #208, opened 2016-08-23
Store docker layers to Travis cache.

glogiotatidis:
Since docker 1.10 and content addressable storage we need to have the
whole chain of layers used to build an image to take advantage of
caching.

Use a specially crafted docker save command to save all layers
explicitly and store them in a Travis cached directory. Before each
build load the cache from the same directory.

See also docker/docker#20316


glogiotatidis:
This is identical to https://github.com/mozmar/lumbergh/pull/178 and tests pass. Therefore merging.

PR #219, opened 2016-10-12
Separate deploys to EUW and USW clusters with tags.

glogiotatidis:
Use different tag format to distinct deploys to EUW and USW clusters.
Tags ending with usw will deploy only to USW and tags ending with euw
will deploy only to EUW.

Also limit deployments to tags matching YYYYMMDD.XX.DDD format where XX
is sequential number of tag for the day starting at 1 (thus 1,2,3...)
and DDD is `usw` or `euw`.

This has been tested against [glogiotatidis/condition-test](https://github.com/glogiotatidis/condition-test/releases). [Travis builds](https://travis-ci.org/glogiotatidis/condition-test/builds)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,13 +71,16 @@ deploy:
     on:
       tags: true
       repo: mozmar/snippets-service
+      condition: $(echo ${TRAVIS_TAG} | egrep "^201[6-9][0-9]{4}.[0-9]+.usw")
   - provider: script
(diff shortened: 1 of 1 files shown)


glogiotatidis:
This will enable us to do rolling deploys for snippets production by tagging twice after manual verification. For staging will be pushing to push clusters at the same time. Is this watch you have in mind @jgmize ?

PR #220, opened 2016-10-12
Use SimpleDictCache (LocMemCache) for product-details

glogiotatidis:


glogiotatidis:
@pmac r?

pmac:
Looks fine to me. Wouldn't be a bad idea to also take [the tests](https://github.com/mozilla/bedrock/blob/master/bedrock/base/tests/test_simple_dict_cache.py), which were mostly taken from Django's test for the LocMem backend. Might make Codecov happier at least :)

r+

glogiotatidis:
Thanks for the code and the review @pmac 🐙

PR #222, opened 2016-12-14
Add campaign based opt outs.

glogiotatidis:
Adds campaign based opt outs, replaces #141

### DIFF
--- a/snippets/base/models.py
+++ b/snippets/base/models.py
@@ -392,7 +392,8 @@ class Snippet(CachingMixin, SnippetBaseModel):
 
     campaign = models.CharField(
         max_length=255, blank=True, default='',
-        help_text='Optional campaign name. Will be added in the stats ping.')
+        help_text=('Optional campaign name. Will be added in the stats ping. '
+                   'Will be used for snippet blocking if set.'))
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
 
--- a/snippets/base/templates/base/includes/snippet.js
+++ b/snippets/base/templates/base/includes/snippet.js
@@ -340,7 +340,8 @@ Mozilla.UITour.setConfiguration = function(configName, configValue) {
         var blockList = getBlockList();
         snippets = snippets.filter(
             function (snippet) {
-                if (blockList.indexOf(snippet.id) === -1) {
+                var blockID = snippet.campaign || snippet.id;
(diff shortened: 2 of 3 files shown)


glogiotatidis:
@ScottDowne please r?

ScottDowne:
Just a heads up I'm still in Hawaii so I'll be doing this review very late
because of time zones.

snippets/base/models.py
```diff
@@ -392,7 +392,8 @@ class Snippet(CachingMixin, SnippetBaseModel):
 
     campaign = models.CharField(
         max_length=255, blank=True, default='',
-        help_text='Optional campaign name. Will be added in the stats ping.')
+        help_text=('Optional campaign name. Will be added in the stats ping. '
+                   'If be used for snippet blocking if set.'))
```
ScottDowne: I think there may be a typo in "If be used for snippet blocking if set."

Not really sure what you were intending though. "To be used for snippet blocking if set."? maybe.

snippets/base/templates/base/includes/snippet.js
```diff
@@ -661,7 +661,9 @@ function popFromBlockList(snippetID) {
 
 function addToBlockList(snippetID) {
     var blockList = getBlockList();
-    snippetID = parseInt(snippetID, 10);
```
ScottDowne: I cannot come up with a good reason.

glogiotatidis: `snippetID` used to be a string (now it's an int) and I parsed it as int so it's faster/better to store and lookup.

With this change we still keep snippetID as int and we also allow strings for campaigns.

snippets/base/models.py
glogiotatidis: I's going to `will be used...` nice catch, fixed

snippets/base/templates/base/includes/snippet.js
ScottDowne: Makes sense to me.

ScottDowne:
@glogiotatidis this looks good to me. Feel free to land this patch.

Just a heads up, I'll be getting on a plane tonight, so I won't be around to help if anything goes wrong, but I also don't have the ability to back anything out here anyway, so I don't think this changes much.

glogiotatidis:
Cool, I'll merge this but hold pushing until monday so we both have stress free weekend. Safe travels back home and thanks for your time :)

PR #223, opened 2016-12-23
Ensure clicks go through, even if metrics url is undefined.

ScottDowne:


ScottDowne:
STR

1. Load up the fundraising snippet.
2. Click donate.

Expected: Goes to the donate page.
Actual: nothing

If you click again, it works.

ScottDowne:
The issue is somehow the snippet-service no longer has a value for `SNIPPET_METRICS_URL`. Without that value, the first attempt to click sets `target.dataset.eventCounted` to true, so it won't fire the metric again.

Then it goes into actually sending the metric, which fails when it sends it nowhere, never actually firing the callback to register the click. If you click again, it bypasses this because `target.dataset.eventCounted` is now set.

Not sure what happened, but that's the root of the issue.

I didn't fix that. I just put in a fail safe so if `SNIPPET_METRICS_URL` is undefined, we can still click things.

ScottDowne:
@glogiotatidis r?

glogiotatidis:
> The issue is somehow the snippet-service no longer has a value for SNIPPET_METRICS_URL

Bundles produced by both EU and US clusters do have `SNIPPET_METRICS_URL` 
 - https://snippets.cdn.mozilla.net/us-west/bundles/bundle_45e76b7585c1832e002ce23ee2b2af433051ce52.html
 - https://snippets.cdn.mozilla.net/eu-west/bundles/bundle_45e76b7585c1832e002ce23ee2b2af433051ce52.html

Also snippets-stats proxy is up and running and GA is receiving data. Trying to reproduce the issue

glogiotatidis:
Managed to replicate this locally. SNIPPET_METRICS_URL is always defined. I believe this is caused by the cyclic call stack of button click. Here's what happens:
 - User clicks donate
 - Template code blocks the click, adds `disabled=disabled` attr, adds to block list, *re-triggers the click*
 - Snippet Service code catches the triggered click, blocks it, calls sendMetric and re-triggers the click

Originally I couldn't reproduce this locally because I had `METRICS_SAMPLE_RATE=1` to always trigger sendMetric pings and therefore the second trigger of the click was always async due to `setTimeout`. 

When I set the sample rate to `0.01` which is the current production setting which mean that most of the time the click is synchronous I could reproduce the issue.

Scott you're right on your analysis that the second click works b/c the `disabled` and `ever-counted` attrs are set and therefore the click goes through uninterrupted. In all cases though it seems the METRICS_URL is defined.

I'll go ahead and merge this as it fixes the problem by the use of `setTimeout` and also adds the extra check metrics url being set.

Many thanks @ScottDowne, great work!

PR #228, opened 2017-04-11
Add extra tags to sentry report.

glogiotatidis:
Extra tags:
 - server_full_name
 - environment
 - site

This those extra tags we will able to merge all Sentry apps into one for easier
managements on multiple clusters.


metadave:
👍

PR #244, opened 2017-07-07
 [bug 1378695] Add Activity Stream startpage option and view. 

glogiotatidis:
Activity stream payload is not final yet.

Includes a django-jinja upgrade to fix `assertTemplateUsed`


glogiotatidis:
Great, thanks! 

I'll keep this unmerged until the next week. I expect the AS team to provide us with a first iteration of the addon with snippets support. Then we can start building the new snippets payload to work with AS.

PR #247, opened 2017-07-18
Add Treeherder link.

glogiotatidis:
@davehunt r? 

Now sure if there's a better badge to add, but thought that a link to Treeherder is a good idea anyway :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ snippets
 
 [Documentation](http://abouthome-snippets-service.readthedocs.org/)
 
+[![webqa tests](https://img.shields.io/badge/WebQA-Treeherder-lightgrey.svg)](https://treeherder.mozilla.org/#/jobs?repo=snippets-service)
+
 Run the tests
 -------------
 


README.md
```diff
@@ -9,6 +9,8 @@ snippets
 
 [Documentation](http://abouthome-snippets-service.readthedocs.org/)
 
+[![webqa tests](https://img.shields.io/badge/WebQA-Treeherder-lightgrey.svg)](https://treeherder.mozilla.org/#/jobs?repo=snippets-service)
```
davehunt: I would suggest treeherder:unknown as these aren't necessarily just web QA tests:

[treeherder](https://treeherder.mozilla.org/#/jobs?repo=snippets-service)

I don't know if Treeherder provides badges, but it would be a nice idea to show the number of jobs for the latest revision and how many of those are passing. Something like:

[treeherder](https://treeherder.mozilla.org/#/jobs?repo=snippets-service)

/cc @edmorley @camd

edmorley: We don't have any badges at the moment - our main use cases are for projects where a pass/fail is sadly often not clear cut (eg intermittents, coalescing, ...) so would be hard to capture that in a single number.

glogiotatidis:
OK I'll merge this as is then and I'm happy to merge a better badge for Treeherder in the future.

Thanks all!

PR #250, opened 2017-07-25
 [fix bug 1376632] Do not preselect channel. 

glogiotatidis:


glogiotatidis:
@pmac please r?

snippets/base/migrations/0019_auto_20170725_1357.py
```diff
@@ -0,0 +1,24 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0018_auto_20170614_0925'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='snippet',
+            name='campaign',
```
pmac: Any idea  why this is here? You don't seem to have altered the `campaign` field.

glogiotatidis: Yeah at some point I changed the help_text. I don't understand why Django monitors those attributes but maybe I'm missing something. Anyway I created another commit with just that to keep this clean. 

https://github.com/mozmeao/snippets-service/commit/fd4f4a061a5dade9ed918ce40d3a09d78ed7cea3

PR #251, opened 2017-07-26
[bug 1378695] Add Activity Stream payload

glogiotatidis:
This is a demo -known to work- payload based on the current Start Page v4
payload. Still needs to be cleaned up and adjusted to AS.


glogiotatidis:
This doesn't change much so I'll self r+.

PR #253, opened 2017-07-27
[bug 1378695] Adjust Activity Stream Payload.

glogiotatidis:
- Remove UITour
 - Remove appInfo, searchengine, fxaccount updaters. Attrs are provided by AS
   directly into gSnippetsMap.
 - Remove gSnippetsMap initialization
 - Remove keypressIsMine workaround (bug 1238092)
 - Adjust addSnippetBlockLinks
 - Remove CDATA since we don't need to be XML valid anymore.
 - Adjust click listener.
 - Remove obsolete code.
 - Remove about:account link handling, since we cannot do it anymore
   without UITour

### DIFF
--- a/snippets/base/templates/base/fetch_snippets_as.jinja
+++ b/snippets/base/templates/base/fetch_snippets_as.jinja
@@ -1,16 +1,13 @@
 <style type="text/css">
   {% include 'base/includes/snippet_as.css' %}
 </style>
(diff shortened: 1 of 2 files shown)


glogiotatidis:
This mostly removes 😆 
image

I'm going to self r+ to get AS testing started. 

/cc @pmac

PR #254, opened 2017-07-27
[bug 1378695] Display #snippets-container only when we have a snippet.

glogiotatidis:
Following up from #253.
See also https://github.com/mozilla/activity-stream/issues/2968


glogiotatidis:
I'm going to self r+ to get AS testing started.

/cc @pmac

PR #255, opened 2017-07-27
Update simple_template_server to work with Activity Stream.

glogiotatidis:


glogiotatidis:
@ScottDowne this is now working and you can start developing snippets on AS!

PR #260, opened 2017-07-31
[bug 1378695] Add block button css to global css.

glogiotatidis:


ScottDowne:
Would it be good to add this too maybe? To match the comps I got?

```
  #snippets-container {
    border-top: solid 2px #E9E9E9;
    background: rgba(255, 255, 255, 0.97);
  }
```

ScottDowne:
I built a bit on top of your pull request. https://github.com/mozmeao/snippets-service/pull/261

Thoughts? The x button is now built in, which is nice!

PR #266, opened 2017-08-14
Don't initialize snippets if onboardingFinished is false.

k88hudson:
In our previous meeting, some concerns were raised about being to override onboarding if necessary. For this reason, I think it would be a good idea to move the logic about whether to show snippets based on onboarding to the service, rather than in Activity Stream (since it would give individual snippets more control)

### DIFF
--- a/snippets/base/templates/base/includes/snippet_as.js
+++ b/snippets/base/templates/base/includes/snippet_as.js
@@ -10,6 +10,11 @@ var GEO_CACHE_DURATION = 1000 * 60 * 60 * 24 * 30; // 30 days
 (function() {
     'use strict';
 
+    // If onboarding hasn't finished its tour, don't load snippets.
+    if (!gSnippetsMap.get('appData.onboardingFinished')) {
+      return;
+    }
+
     if (ABOUTHOME_SNIPPETS.length > 0) {
         ABOUTHOME_SHOWN_SNIPPET = chooseSnippet(ABOUTHOME_SNIPPETS);
     }


bensternthal:
@pmac is this something you feel qualified to review/merge? If not we will need to wait until Giorgos returns from PTO next week.

pmac:
The change is minimal and looks syntactically correct, but I'm not really in a position to say whether this is a good idea or the proper way to go about it. I'm fairly confident it's good, but if we can wait for Giorgos I'd feel a lot better about it.

k88hudson:
I think this is fine to wait until @glogiotatidis gets back 👍

glogiotatidis:
> In our previous meeting, some concerns were raised about being to override onboarding if necessary. For this reason, I think it would be a good idea to move the logic about whether to show snippets based on onboarding to the service, rather than in Activity Stream (since it would give individual snippets more control)

I'm not entirely sure I understand how override will work in this case. Onboarding display is still controlled by AS. The way I understand this patch is that we can get to ignore `gSnippetsMap.get('appData.onboardingFinished')` on demand and thus force display snippets on top of onboarding. My concerns are the following:

 - The onboarding fox on the upper left corner will still appear
 - The experience may be sub-optimal if we first draw onboard and then display snippets or even possibly we can get ourselves into a race condition.

@k88hudson can you please provide some insight?

k88hudson:
@glogiotatidis What I was thinking is that I could add a method to `gSnippetsMap` to trigger the override, this PR is just the first step to make it possible (if the snippets payload never loads at all it can't do any decision-making). Although you bring up a good point with the race condition – we'll have to figure out how to handle that. Maybe it would only trigger the override on the next visible new tab?

k88hudson:
The ideal situation of course would be  for onboarding to use the same decision-making mechanism as the rest of snippets, and the snippets service could decide what to do, but I'm assuming we're not going to get that for 57

glogiotatidis:
> @glogiotatidis What I was thinking is that I could add a method to gSnippetsMap to trigger the override, this PR is just the first step to make it possible (if the snippets payload never loads at all it can't do any decision-making). Although you bring up a good point with the race condition – we'll have to figure out how to handle that. Maybe it would only trigger the override on the next visible new tab?

Can we call onBoarding from snippets.js somehow? Something like `showOnboarding()`? If so we can change the code to something like this:

https://github.com/k88hudson/snippets-service/blob/6b8e5cce10885cdd251f3979ebd40872635b387e/snippets/base/templates/base/includes/snippet_as.js#L18-L20

```js
if (ABOUTHOME_SNIPPETS.length > 0) {
  ABOUTHOME_SHOWN_SNIPPET = chooseSnippet(ABOUTHOME_SNIPPETS);
  if (!ABOUTHOME_SHOWN_SNIPPET.overrideOnboarding && !gSnippetsMap.get('appData.onboardingFinished')) {
    showOnboarding();
    return;
  }
}
```

With this code we can have individual snippets that can override the on-boarding snippets.

k88hudson:
@glogiotatidis that sounds like the right way to go, yes. We'll need to work with the onboarding folks to figure this out, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1393489 to discuss

PR #269, opened 2017-08-23
Upgrade to Django 1.11

glogiotatidis:
This PR upgrades to the latest LTS Django (1.11). 

Decisions:
 - Remove `SimpleDictCache` used to cache `django-product-details` dictionaries due to failing django tests. It gets replaced by simple `LocMemCache`. 
 @pmac I originally copied the `SimpleDictCache` code from bedrock. Do you use it there because of the faster setting / fetching due to the lack of pickling? Do you have any concerns for replacing it with SimpleDictCache?

 - Removed the deprecated `django-cache-machine`.  I could have replaced that with something like `django-cacheops` but I'm working on an overall new caching plan which changes things. 

To make sure that the two changes above don't bring down the app I'll push to one cluster only and see who it behaves with real traffic before going global.

 - Switching from `django-session-csrf` to django's CSRF_USE_SESSIONS

Other notes:
 - Removed unused `active-snippets` view
 - Removed unused bcrypt (from the pre-saml era)
 - Upgrade to debian:stretch


snippets/base/urls.py
```diff
@@ -18,7 +18,6 @@
     url(r'^preview/$', views.preview_snippet, name='base.preview'),
     url(r'^show/(?P<snippet_id>\d+)/$', views.show_snippet, name='base.show'),
     url(r'^json-snippets/', views.JSONSnippetIndexView.as_view(), name='base.index_json'),
-    url(r'^active-snippets.json', views.ActiveSnippetsView.as_view(), name='base.active_snippets'),
```
pmac: is this just no longer in use?

snippets/settings.py
```diff
@@ -223,11 +218,11 @@ def mysql_version(self):
 CACHES = {
     'default': config('CACHE_URL', default='locmem://', cast=django_cache_url.parse),
     'product-details': {
-        'BACKEND': 'snippets.base.cache.SimpleDictCache',
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
         'LOCATION': 'product-details',
         'OPTIONS': {
-            'MAX_ENTRIES': 200,  # currently 104 json files
-            'CULL_FREQUENCY': 4,  # 1/4 entries deleted if max reached
+            'MAX_ENTRIES':   200,  # currently 104 json files
```
pmac: NIT: Looks like a couple extra spaces were added here.

snippets/urls.py
```diff
@@ -2,6 +2,7 @@
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.contrib.staticfiles.urls import staticfiles_urlpatterns
+
```
pmac: NIT: superfluous extra line

snippets/base/urls.py
glogiotatidis: Yes, it was a request for a custom view back in the day. I crossed checked with the original consumers and they don't use it. We never advertised it either.

glogiotatidis:
Thanks for your time @pmac

PR #273, opened 2017-08-29
Don't expire the bundle key in the cache.

glogiotatidis:
The snippet bundle cache key is calculated (also) based on the snippet.modified value. So if a snippet to be included in the bundle gets modified the key will and thus a new bundle will be generated. There 's no need to expire the bundles after `settings.SNIPPET_BUNDLE_TIMEOUT` minutes. Instead let them get evicted from the cached and replaced by newer bundle key.

This should reduce the CPU usage of the snippets-service and provide the exact same experience.

Tested on Tokyo and Frankfurt with real traffic and the verified that bundles do get generated only when a snippet changes.


glogiotatidis:
@pmac could you please r?

glogiotatidis:
Thanks!

PR #276, opened 2017-09-01
Cache Database queries in LocMem for 5 minutes with django-cachalot.

glogiotatidis:
This will reduce the stress on MariaDB and will not increase the traffic to
Redis in the cost of a little more memory usage per process.

The fact the LocMem is not shared between processes does not raise any issues
except that we use more local memory. DB remains the only source of truth and
the cache will get invalidated every 5 minutes to keep up to date with it.


glogiotatidis:
@pmac this comes as a replacement to django-cache-machine with the difference that it caches locally.

glogiotatidis:
Thanks!

PR #279, opened 2017-09-11
Store bundles in cache instead of in S3.

glogiotatidis:
- Store bundles into cache (Redis in production) instead of S3.
- Merge fetch_pregenerated_snippets and fetch_render_snippets into one.

With this change the code gets simplified and big performance gains are expected
by removing the App -> S3 and CDN -> App -> S3 trips.

Also workers running in different regions will now return the same CDN URL
instead of a region specific URL

  New: https://snippets.cdn.mozilla.org/4/...
  Old: https://snippets.cdn.mozilla.org/us-west/bundles/bundle_....

which should improve the cache hit rate and distribute the load to different
regions better.

Added Etag support for the bundled snippets response.

Infra provisioning is simpler without the need of S3, except for the Admin App
which needs access to deploy files.

Tested on Tokyo for a short time and works as expected!

Before deploying:
 - [ ] Unset `SNIPPET_HTTP_MAX_AGE` from deis apps to get default value.
 - [ ] Unset `AWS_*` env variables from all hosts except from `snippets-admin`
 - [ ] Unset `FILE_STORAGE` env variable from all hosts except from `snippets-admin`
 - [ ] Update mozmeao/infra with new variables

### DIFF
--- a/snippets/base/models.py
+++ b/snippets/base/models.py
@@ -14,8 +14,6 @@ from xml.sax import ContentHandler
 from django.conf import settings
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
-from django.core.files.base import ContentFile
-from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models.manager import Manager
@@ -135,6 +133,7 @@ class SnippetBundle(object):
     def __init__(self, client):
         self.client = client
         self._snippets = None
+        self._contents = None
 
     @property
     def key(self):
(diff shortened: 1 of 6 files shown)


glogiotatidis:
> I assume that the new CDN bundle URLs will simply use the app as the origin?
yes

> I'm still a little hazy on why it's better to serve the bundles from the app (cache or no) than from flat files in s3, but I trust your judgement on that.

So I believe that three things will make a difference:
 - App will be faster b/c will not write to S3 (but, also slower b/c will serve bundle itself)
 - Bundle URLs will be the same across datacenters and therefore the load will be distributed better.
 - Clients will not have to do 2 roundtrips, one to hit the App and then the S3 bucket. 

I'll run the experiment and share the results with the team. Thanks!

PR #287, opened 2017-09-14
Add oidc support

glogiotatidis:
- Reverts the SAML/Okta commit
- Add oidc support

Tested against stage with huge success.


snippets/base/templates/admin/login.html
```diff
@@ -0,0 +1,62 @@
...
+            <a href="{{ password_reset_url }}">{% trans 'Forgotten your password or username?' %}</a>
+          </div>
+        {% endif %}
+        <div class="submit-row">
+          <label>&nbsp;</label><input type="submit" value="{% trans 'Log in' %}" />
+        </div>
+      </form>
+    {% endif %}
+    <script type="text/javascript">
+     document.getElementById('id_username').focus()
```
pmac: I'm sure this is just standard Django login form, but instead of inline JS we could just use the [autofocus HTML5 attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autofocus).

Not a blocker, just a suggestion.

glogiotatidis: Indeed that's the standard django admin login page. I'd like to implement autofocus but I guess we need to either manual write the html for username input field or alter the widget on the django level. 

Is there a simpler build-in way to archive this?

glogiotatidis: BTW I'm going to merge this today as is to get oidc live but I don't mean to stop this discussion.

glogiotatidis:
> Looks great! Nice work. I'll be stealing this for basket and probably Nucleus :) 💯

Yes, django-mozilla-oidc is very nice implementation and integration. Super easy to implement too.

snippets/base/templates/admin/login.html
pmac: Cool. I thought I remembered a template filter that would add an attribute to a django field, but now I can't find it. This is certainly not a big deal, but it would be nice :) Especially if you want to turn on CSP at some point.

glogiotatidis: Good point on CSP, I took a TODO to fix this. Thanks!

PR #294, opened 2017-09-28
Custom WSGI request_class for manual HTTPS setting.

glogiotatidis:


glogiotatidis:
We're not going to use this since we switched elbs to http/https

PR #295, opened 2017-09-29
Update django-revertion and add support to SnippetTemplate

glogiotatidis:
- [ ] After merging run `./manage.py createinitialrevisions`


glogiotatidis:
Thanks!

PR #296, opened 2017-10-04
Merge S3 buckets 

glogiotatidis:
Currently we're using different buckets and bucket locations for bundles depending on the region. This has some downsides:

 - Makes infrastructure and deployments of new cluster more complex. 
 - Reduces the benefits of CDN since different regions return different URLs (due to the different S3 origins) for the same content.

The upside of having the buckets geographically close to each cluster is lower upload times but since 35128394c7cfe88875fdc79fa9fd3ff6bb9bfcf1 we don't generate that many bundles.

This PR checks both Redis cache (first) and S3 (second) before generating a new bundle. If a bundle exists in S3 but not in Redis adds the key to the cache. 

It also reduces the key caching time down to one day to handle the auto-deletion of bundles in S3 better.

Need to change the following settings in `tokyo` and `frankfurt` after deploying to match `us-west`.

 - `AWS_STORAGE_BUCKET_NAME`
 - `AWS_S3_HOST`
 - `MEDIA_BUNDLES_ROOT`

Also run `./manage.py clear_cache` in 

 - [ ] tokyo
 - [ ] frankfurt 

after pushing.

### DIFF
--- a/snippets/base/models.py
+++ b/snippets/base/models.py
@@ -35,6 +35,8 @@ from snippets.base.managers import ClientMatchRuleManager, SnippetManager
 from snippets.base.util import hashfile
 
 
+ONE_DAY = 60 * 60 * 24
+
 JINJA_ENV = engines['backend']
 
 SNIPPET_JS_TEMPLATE_HASH = hashfile(
@@ -225,7 +227,7 @@ class SnippetBundle(object):
         if isinstance(bundle_content, unicode):
             bundle_content = bundle_content.encode('utf-8')
         default_storage.save(self.filename, ContentFile(bundle_content))
-        cache.set(self.cache_key, True, None)
+        cache.set(self.cache_key, True, ONE_DAY)
 
 
 class SnippetTemplate(models.Model):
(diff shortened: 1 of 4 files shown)


snippets/base/models.py
```diff
@@ -135,13 +137,17 @@ class SnippetBundle(object):
...
         self._snippets = None
+        self._key = None
 
     @property
     def key(self):
         """A unique key for this bundle as a sha1 hexdigest."""
         # Key should consist of snippets that are in the bundle plus any
         # properties of the client that may change the snippet code
         # being sent.
+        if self._key:
```
pmac: This works fine, but you could also use [Django's `@cached_property` decorator](https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.functional.cached_property) for these kinds of methods. It basically runs the function the first time, then alters the class so that the method is replaced by a simple property with the value of the function's return. Not a necessary change at all, but cool I think for future reference.

glogiotatidis: Useful, thanks!

PR #297, opened 2017-10-04
Remove bleach.

glogiotatidis:
bleach is not used and this version in insecure.


glogiotatidis:
Simple change, tests pass, self r+

PR #300, opened 2017-10-16
[fix bug 1408960] Make snippet weight editable in list view.

glogiotatidis:


glogiotatidis:
> Probably didn't really need a review on this one, but it LGTM :) 👍

True but you usually have nice django tips :) Thanks!

PR #303, opened 2017-10-18
[bug 1408981] Use more  values and settings to calculate bundle.key

glogiotatidis:


glogiotatidis:
Broke it to multiple commits to make it easier to review but it's all about using more variables to calculate `bundle.key`

glogiotatidis:
Thanks!

PR #304, opened 2017-10-19
 [fix bug 1408968] Fix snippet size check when saving snippets in Admin 

glogiotatidis:
Fixing the bug and slight refactor.


glogiotatidis:
Thanks!

PR #308, opened 2017-10-25
[fix bug 1408961] Allow preview of Snippets for non-Admin users

glogiotatidis:


pmac:
Why the need for the new UUID field? Can't there just be a read-only preview URL using the PK?

glogiotatidis:
> Why the need for the new UUID field? Can't there just be a read-only preview URL using the PK?

Good question. We already have the preview URL using the PK. This URL will display the snippet if the snippet is published (== disabled=False) or the user is logged in. 

What we're trying to solve here is the case where the snippet un-published and the editors want to share it with users who do not have a login. Thus the use of the UUID and the "secret" preview URL

pmac:
I guess I just didn't realize the other idea was a secret. Is there some problem of sharing the PK with a non-logged in user? Otherwise we could just make a new URL that allowed them to see the snippet by ID. Is the problem that if we did that with sequential IDs they'd be able to just see all of them by guessing the URLs? That seems like a problem, and uuid4s are a good way of solving it. Thanks!

glogiotatidis:
Good point on keeping the 2 commits. Thanks!

PR #310, opened 2017-10-26
[Fix Bug 1411921] Auto disable snippets that have passed their publish end date.

glogiotatidis:
Add 
 - [x] DEAD_MANS_SNITCH_PRODUCT_DETAILS to snippets-admin
 - [x] DEAD_MANS_SNITCH_PRODUCT_DETAILS to snippets-stage
 - [x] DEAD_MANS_SNITCH_DISABLE_SNIPPETS to snippets-admin
 - [x] DEAD_MANS_SNITCH_DISABLE_SNIPPETS to snippets-admin

Remove after pushing to prod DEAD_MANS_SNITCH_URL
 - [ ] snippets-stage
 - [ ] snippets-admin


scripts/cron.py
```diff
@@ -70,6 +58,15 @@ def job_update_product_details():
     connections.close_all()
 
 
+@scheduled_job('cron', month='*', day='*', hour='*', minute='*', max_instances=1, coalesce=True)
+@babis.decorator(ping_after=settings.DEAD_MANS_SNITCH_DISABLE_SNIPPETS)
+def job_disable_snippets_past_publish_date():
+    call_command('disable_snippets_past_publish_date')
+    # Django won't close db connections after call_command. Close them manually
```
pmac: Just a note: turns out a lot of these management commands assume they're not in a long-running process, so when doing this in bedrock I just [shell out and call the management command as a separate process](https://github.com/mozilla/bedrock/blob/d26bad24956a107a76ff5da6bb9297c6d1747457/bin/cron.py#L25-L26). It's not that much more time and avoids some issues like this. Not an r-, just something to keep in mind if more strange behavior is found.

glogiotatidis: Actually this is good advice and I'm going to take it.

glogiotatidis:
Updated with call_command and also checked the new job to run every 12 hours instead of every minute.

PR #315, opened 2017-11-01
 [fix bug 1408986] Manual ordering for Template Variables. 

glogiotatidis:
This is a simple way to enable us to manual order the template variables so we can group them together by context.

In the future we can explore parsing the starting template comment ([example](https://github.com/mozmeao/snippets/blob/master/activity-stream/simple-snippet.html#L1-L15)) and extract from it the order and also the type and description.


glogiotatidis:
Thanks!

PR #320, opened 2017-11-08
Switch docker base image to python:2.7-slim

glogiotatidis:
Re https://github.com/mozmeao/infra/issues/641

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM python:2.7-slim-stretch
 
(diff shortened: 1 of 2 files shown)


glogiotatidis:
> LGTM. To be clear though I think debian:stretch is a fine base image and definitely well-maintained. If you wanted to stick with it I wouldn't fault you. I think the official Python image has a more updated Python, but that's really a preference.

Agree. Up-to-date python sounds like a good idea and i like common patterns in projects. Thanks pmac!

PR #333, opened 2017-12-07
Hiding x button until snippet is hovered over.

ScottDowne:


ScottDowne:
@glogiotatidis ping on this.

Jean asked if I could create a small patch to hide the x until hover.

glogiotatidis:
I remember that this was explicitly made to always show, after having it to show only on hover. Is this change permanent or an experiment?

ScottDowne:
Jean didn't mention a test. We should double check.

jkt230:
Thanks @ScottDowne - @glogiotatidis The fundraising, Product (and UX) and Lifecycle marketing teams all got together to figure out ways to improve conversion and perception of the Activity Stream snippets (based on engagement and survey results). One of the ways to see if we can improve results is by making the X only appear when it hovers because right now having the X always show kind of invites users to X out right away. 

I've filed a bug for next year with UX team to see if we can design a flow where if a user X out three or more snippets in a row that they'll be shown a prompt that they can turn off snippets in the preference center.

ScottDowne:
@jkt230 thanks, one last point I want to be crystal clear on, did you want this x change across all activity strea snippets, or just fundraising ones?

jkt230:
@ScottDowne on all please, the experience should be consistent.

glogiotatidis:
In prod!

PR #361, opened 2018-02-21
Add django-allow-cidr

glogiotatidis:


glogiotatidis:
Blocked by https://github.com/mozmeao/django-allow-cidr/pull/1

glogiotatidis:
This is straight forward and we discussed in other prs. tests pass so self merging.

PR #364, opened 2018-03-02
Extent profile age weeks options from 8 to 17 weeks.

glogiotatidis:


glogiotatidis:
Simple addition of options, self merging.

PR #365, opened 2018-03-06
Fix Bug 1438368 - Call blockSnippet if it exists

piatra:
Added a new method for blocking snippets that should be called if serving Firefox 60 or newer.
The old `gSnippetsMap.set('blockList', ...)` still works as expected and calling both is not an issue.

Activity Stream PR is here https://github.com/mozilla/activity-stream/pull/4014

### DIFF
--- a/snippets/base/templates/base/includes/snippet_as.js
+++ b/snippets/base/templates/base/includes/snippet_as.js
@@ -517,7 +517,7 @@ function addToBlockList(snippetID) {
     }
     if (blockList.indexOf(snippetID) === -1) {
         blockList = [snippetID].concat(blockList);
-        gSnippetsMap.set('blockList', blockList);
+        gSnippetsMap.blockSnippetById(snippetID);
     }
 }
 


snippets/base/templates/base/includes/snippet_as.js
```diff
@@ -518,6 +518,9 @@ function addToBlockList(snippetID) {
     if (blockList.indexOf(snippetID) === -1) {
         blockList = [snippetID].concat(blockList);
         gSnippetsMap.set('blockList', blockList);
+        if (gSnippetsMap.blockSnippet) {
+          gSnippetsMap.blockSnippet(snippetID);
+        }
```
k88hudson: this should be an else `gSnippetsMap.set('blockList', blockList);` instead of that being above;
also, can you add a comment with a link to the bug number explaining this? thanks!

piatra:
This needs to land before https://github.com/mozilla/activity-stream/pull/4014.

piatra:
@glogiotatidis hey, PR landed in m-c this should land as well.

glogiotatidis:
Excellent! this is now live.

PR #375, opened 2018-04-11
Django1.11.12

glogiotatidis:


glogiotatidis:
Simple django upgrade, tests pass, self-merging

PR #384, opened 2018-04-23
 [Fix #383] Enable brotli compression of bundles. 

### LINKED ISSUE #383: Enable Brotli compression of bundles.

glogiotatidis:
Upload pre-compressed bundles to S3 using the brotli compress method which achieves significantly better compress ratios for snippet content compared to gzip.

- To enable set `BUNDLE_BROTLI_COMPRESS` to True
- Only serves brotli content to activity stream clients which are guarantied to be at least version 57 (brotli was introduced in Fx 44)
- I tested with CF and it continues to automatically gzip non-brotli compressed files, while serving br files with correct metadata. So no change needed for the CDN.


glogiotatidis:
Thanks!

PR #392, opened 2018-04-26
[Fix #341] Make Activity Stream the default in start page version sel…

### LINKED ISSUE #341: Make 'Activity Stream' the default start page option for new snippets
Activity stream is the new cool kid on the block. Let's make all the new snippets default to that Startpage instead of the current Versions 2,3,4.

image

glogiotatidis:
…ection.


glogiotatidis:
> Code does look good but there are test failures.




 fixing!

PR #393, opened 2018-04-26
[Fix #346] Add filter for time since last session

### LINKED ISSUE #346: [new filter] Filter based on date of last Firefox use.
Currently waiting for information to become available. [Bug 1431469](https://bugzilla.mozilla.org/show_bug.cgi?id=1431469)

pmac:
I couldn't really tell from the bugs but it looked like they wanted something for this exactly like the profile age stuff. If "weeks" is not a granular enough time slice we could switch to days if they'd like.


pmac:
Doh! Failed to make a branch for this. Opening a new one.

PR #394, opened 2018-04-26
[Fix #346] Add filter for time since last session

### LINKED ISSUE #346: [new filter] Filter based on date of last Firefox use.

pmac:
I couldn't really tell from the bugs but it looked like they wanted something for this exactly like the profile age stuff. If "weeks" is not a granular enough time slice we could switch to days if they'd like.

### DIFF
--- a/snippets/base/admin.py
+++ b/snippets/base/admin.py
@@ -210,6 +210,8 @@ class SnippetAdmin(BaseSnippetAdmin):
                 'client_option_is_developer',
                 'client_option_is_default_browser',
                 'client_option_screen_resolutions',
+                'client_option_sessionage_lower_bound',
+                'client_option_sessionage_upper_bound',
                 'client_option_profileage_lower_bound',
                 'client_option_profileage_upper_bound',
             )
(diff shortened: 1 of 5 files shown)


glogiotatidis:
> I couldn't really tell from the bugs but it looked like they wanted something for this exactly like the profile age stuff. If "weeks" is not a granular enough time slice we could switch to days if they'd like.

You're right. They want something to cover
```
- Snippets are only displayed to users based on the last session end date targeting used
-- Examples:
---1. Display snippet to users whose last session (prior to the current one) ended at least 4 weeks ago
---2. Display snippet to users whose last session (prior to the current one) ended at least 52 weeks ago
---3. Display snippet to users whose last session (prior to the current one) ended at least 4 weeks ago but no more than 8 weeks ago```

so we're good

glogiotatidis:
Merged https://github.com/mozmeao/snippets-service/commit/d4476f57d91e9c691d4bcf996622f43216b71eb5

PR #501, opened 2018-05-16
[fix #340] Allow main text editing from list view.

### LINKED ISSUE #340: Enable Main Text editing from the list view.
We want to be able to alter the Main Text of the snippet from the admin list view of snippets. 

Since `Main Text` is a SnippetTemplateVariable object and not a Snippet attribute this is a bit more involved. 

It may also require that we restrict templates to have only one `Main Text` variable, which reflects the reality btw: No used template has more than one `Main Text` variables.

The UX/UI can be interesting too, it's suggested that we start small and we iterate with the Snippet Editors to get the best experience.

image

glogiotatidis:
Allow main text editing from list view.

This uses `django-quickedit-ng` to allow drop down editing like this

image

@pmac I had to fork [django-quickedit](https://github.com/callowayproject/django-quickedit/) to make it work with latest django versions. If you don't mind take a look at [this code](https://github.com/glogiotatidis/django-quickedit-ng) as well.


snippets/base/forms.py
```diff
@@ -148,6 +149,32 @@ class BaseSnippetAdminForm(forms.ModelForm):
     pass
 
 
+class SnippetChangeListForm(forms.ModelForm):
+    class Meta:
+        model = Snippet
+        fields = ('body',)
```
pmac: Nit: blank line after the internal class

snippets/base/admin.py
```diff
@@ -167,6 +170,8 @@ def locale_list(self, obj):
 
 
 class SnippetAdmin(BaseSnippetAdmin):
+    def get_changelist_form(self, request, **kwargs):
+        return forms.SnippetChangeListForm
```
pmac: Nit: blank line after the method

snippets/base/forms.py
glogiotatidis: /me wonders why flake8 doesn't complain about this?

glogiotatidis:
Thanks @pmac! :smile:

PR #513, opened 2018-05-21
Drop disabled column

glogiotatidis:


glogiotatidis:
Replaced by #604 and #605 to allow graceful deployments

Issue #353, opened 2018-02-12
Enable more permissions/roles so non-admin users can build snippets but not publish


jkt230:
To streamline the IPC process, we want to enable content writers to build snippets within the service for review

glogiotatidis:
note for when we start building: This requires more detailed permissions which are described in the IPC 2018 requirements document.

glogiotatidis:
From Snippets Service [Product Requirements doc](https://docs.google.com/a/mozilla.com/document/d/12bCVSEQnnTkvrLHzMh3yGY7taUE7KFYZpiZ7WL2F_q0/edit?disco=AAAABi1f-l0)

"""
Giorgos:
Currently we have 2 level permissions:

1. Users can add/edit/delete/publish Desktop Snippets
2. Users can add/edit/delete/publish
Android Snippets

We can add more so that

1. Users can add/edit/delete all desktop
2. Users can publish all desktop
3. Users can add/edit/delete only snippets they have created
4. Users can publish only theirs
5. User can publish only on nightly

etc

Can you please make a list of the must have and nice to have permissions?
Show less


Jean Collings
8:39 PM Jan 12
I imagine that for each channel (release, beta, nightly, dev edition) there would be:
1) Viewer (who can only view)
2) Creator (who can view and draft - add/edit/delete)
3) Publisher/Admin (who can view, draft and publish)


Giorgos Logiotatidis
9:29 AM Jan 15
1) Viewer views all snippets (disabled or otherwise) or just active snippets?

Jean Collings
8:25 PM Jan 17
1) view all snippets please
"""

glogiotatidis:
I removed my assignment b/c probably i won't get to this before may 7th. @pmac feel free to take it on if you have capacity

glogiotatidis:
Interesting projects:
 - https://github.com/silentsokolov/django-admin-permissions
 - https://github.com/lukaszb/django-guardian
 - https://github.com/dfunckt/django-rules

glogiotatidis:
Created: 
 - #534 
 - #535

glogiotatidis:
Closing this one in favor of the newly created ones.

PR #524, opened 2018-05-23
[Issue #353] Publish button

glogiotatidis:
Convert `published` from a check-box to a button with confirmation pop-up before publishing. 

image

Will extend with permission checking

### DIFF
--- a/snippets/base/static/js/admin-publish.js
+++ b/snippets/base/static/js/admin-publish.js
@@ -1,0 +1,11 @@
+django.jQuery(document).ready(function() {
+    django.jQuery('[data-tool-name="publish_object"]').on("click", function(event) {
+        function confirmation() {
+            let answer = confirm("Really publish Snippet?");
+            if (!answer) {
+                event.preventDefault();
+            }
+        }
+        confirmation();
+    });
+});
--- a/snippets/base/tests/test_admin.py
+++ b/snippets/base/tests/test_admin.py
@@ -3,47 +3,35 @@ from django.test.client import RequestFactory
 
 from mock import Mock, patch
 
-from snippets.base.admin import SnippetAdmin, SnippetTemplateAdmin
-from snippets.base.models import Snippet, SnippetTemplate, SnippetTemplateVariable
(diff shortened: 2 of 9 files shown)


snippets/base/static/js/admin-publish.js
```diff
@@ -0,0 +1,11 @@
+django.jQuery(document).ready(function() {
+    django.jQuery('[data-tool-name="publish_object"]').on("click", function(event) {
+        function confirmation() {
```
pmac: For my own curiosity: why did you need to put this in a nested function and then immediately call it?

glogiotatidis: because i suck at javascript? :D

snippets/base/tests/test_admin.py
```diff
@@ -3,47 +3,35 @@
...
+        User.objects.create_superuser(username='admin',
+                                      email='foo@example.com',
+                                      password='admin')
+        self.client.login(username='admin', password='admin')
+
+    def test_action_publish_snippet(self):
+        snippet = SnippetFactory(published=False)
+        self.assertFalse(snippet.published)
+        snippet_action_url = '/admin/base/snippet/{}/actions/publish_object/'.format(snippet.pk,)
+        self.client.get(snippet_action_url)
```
pmac: Hmm.. didn't notice this before, but it's odd that the action can be taken using a GET request. Should really be a POST to modify an object.

glogiotatidis: That's a very good point and I'm immediately returning this to WIP. It seems that this is how django-object-actions implements things. I'll re-think the approach

glogiotatidis:
I'm going to close this one as I don't like the approach and while related it's not a blocker to tackle #353 

Thanks for the insightful info @pmac

pmac:
That's a bummer. I was thinking how useful a project object-actions was. I don't think it would take a huge change to fix this GET/POST issue, but I could easily be wrong.

glogiotatidis:
> That's a bummer. I was thinking how useful a project object-actions was. I don't think it would take a huge change to fix this GET/POST issue, but I could easily be wrong.

They have a couple related issues open, maybe I'll take a look later on.

Issue #536, opened 2018-06-04
[new permission] Snippet Publisher


glogiotatidis:
For each channel (release, beta, etc) create a new permission that will allow users to publish snippets.

 - A user with permission to publish on X channel must be allowed to set `published` to `True` from `False` and the other way around when the snippet is set to publish on X channel. 
 - If multiple publication channels are selected, user must have publication permissions for all selected channels. 
 - A user without permission on all publication channels must be forbidden to set `published` to `True` from `False` or edit any parameter of the published snippet.

This is currently only for the Desktop Snippets 

- [ ] Assign permissions via the admin once this gets pushed out to prod

glogiotatidis:
@ldwright updated the [permission model](https://github.com/mozmeao/snippets-service/issues/536#issue-329124397). I ended up implementing more granular per-channel permissions so we can allow users to have pre-release only publication rights.

PR #537, opened 2018-06-04
[Issue #536] Create snippet publisher permission and validation.

glogiotatidis:
Don't allow users to assign snippets to channels unless they have the
corresponding can_publish_on_<channel> permission.


snippets/base/forms.py
```diff
@@ -350,6 +350,13 @@ def clean(self):
              cleaned_data.get('client_option_addon_check_type') != 'any')):
             raise forms.ValidationError('Type add-on name to check or remove add-on check.')
 
+        # Check publish permissions per channel.
+        for channel in CHANNELS:
```
pmac: Should this check only apply when `published` is `True`, or are these permissions designed more to prevent people from editing any snippet on a channel they don't have the publish permission for?

glogiotatidis:
@pmac updated the logic to apply only when `published=True` as you suggested. You may want to review [permission model](https://github.com/mozmeao/snippets-service/issues/536#issue-329124397) that I'm implementing here.

Can you please re-review?

glogiotatidis:
> The checks don't seem to apply when clicking the "published" check box on the list view. i.e. a user that doesn't have permissions for release channel can mark a snippet published for release channel from the list view but not from the snippet form view.

Fixed this one. UX is still an issue, we need to escape django admin.

@pmac r?

pmac:
> Fixed this one. UX is still an issue, we need to escape django admin.

Looks good! Also agreed. r+

PR #601, opened 2018-07-17
Force close DB connection to init a new one.

glogiotatidis:
This will fix the MySQL has gone away errors.

Verified working on stage.

### DIFF
--- a/scripts/cron.py
+++ b/scripts/cron.py
@@ -5,6 +5,7 @@ import sys
 from subprocess import check_call
 
 from django.conf import settings
+from django.db import connection
 
 import babis
 from apscheduler.schedulers.blocking import BlockingScheduler
@@ -55,6 +56,7 @@ class scheduled_job(object):
 @babis.decorator(ping_after=settings.DEAD_MANS_SNITCH_PRODUCT_DETAILS)
 def job_update_product_details():
     call_command('update_product_details')
+    connection.close()
     create_countries()
     create_locales()
 


scripts/cron.py
```diff
@@ -55,6 +56,7 @@ def log(self, message):
 @babis.decorator(ping_after=settings.DEAD_MANS_SNITCH_PRODUCT_DETAILS)
 def job_update_product_details():
     call_command('update_product_details')
+    connection.close()
```
pmac: This is very interesting. So the data connection issue is happening in the cron.py process and not the management command itself? It's interesting that this process connects to the database at all... perhaps we could manually setup Django here to just connect to an in-memory sqlite3 db and avoid this connection all together?

pmac: You can override Django setup and settings similar to how it's done here:

https://github.com/mozmeao/django-allow-cidr/blob/master/runtests.py#L18

In those alternate settings you could set the `DATABASES` setting to use the SQLite3 db backend and `:memory:` for the filename and it will create and use an in-memory database, which is the fastest and least impactful one which is great for this since it doesn't need a DB at all.

pmac: I guess in this case we do want the real settings, so we could just have the DB connection be configurable and set `DATABASE_URL=sqlite://:memory:`.

glogiotatidis: I understand that since `cron.py` is run using `runscript` the whole Django stack -including the db connection- is setup. In this case we do actually need to have a connection to the real database for the commands following the `update_product_details` cmd.

What's happening here should be the following:
 - `runscript` setups Django and connects to the database. This is a long running command, so at some point the database connection times out.
 -  When it's time to run the `job_update_product_details` job the connection is already timed out and the second part of the job (after the management command) fails with the "MySQL has gone way" error.

Closing the connection should fix this problem. Alternatively we can move the `create_countries` and `create_locales` to their own management command and call it after the `update_product_details` cmd

pmac: Hmm. Right. Didn't notice that those functions were being called directly and not in separate processes like the management command. If this works I say ship-it.

PR #612, opened 2018-07-20
Change pyup schedule to biweekly.

glogiotatidis:


glogiotatidis:
Also emailed pyup to ask about how they handle sec updates in this case. 

/cc @pmac

PR #658, opened 2018-08-16
Django 2.1

glogiotatidis:
- Upgrade to Django 2.1
- Removed django-filters and pages listing published snippets on `/` b/c there's no-one watching those now. When rebuild we will use another filtering method with JEXL
- Removed admin advanced filters b/c they are not used.


snippets/base/urls.py
```diff
@@ -7,25 +7,18 @@
 
 
 urlpatterns = [
-    url(r'^$', views.SnippetIndexView.as_view(), name='base.index'),
-    url(r'^(?P<startpage_version>[^/]+)/(?P<name>[^/]+)/(?P<version>[^/]+)/'
```
glogiotatidis: loving how you can cast url variables now. Here making `startpage_version` an int.

jpetto: Oooh!

pmac: Ooooooooo! So this is something the new `django.urls.path` can do? I need to read up on this. Looks nice! A bit more Flasky.

pmac: Oh yes. It's very nice. https://docs.djangoproject.com/en/2.1/topics/http/urls/#path-converters

Need to get a project up to Django 2.0+ :)

glogiotatidis:
Updates

pmac:
Looks like your build on Travis got bit by https://github.com/redis/hiredis-py/issues/38. They recently uploaded a bunch of wheels for `hiredis` which breaks our hashes. Just need to `hashin` again to get the new hashes.

glogiotatidis:
Thanks pmac!

PR #662, opened 2018-08-21
Remove FetchSnippetsMiddleware

glogiotatidis:
With the Django 2 upgrade it appears that django.urls.resolve used by
FetchSnippetsMiddleware has been randomly raising 500s, with CSRF Middleware
complaining that request.session is not defined.

I've managed to replicate this issue by using DEBUG=False and ./bin/run-prod.sh
locally, after cleaning db sessions and local cookies. Sometimes the error
doesn't occur though, probably when other pages are visited before and Django's
URL cache gets bootstrapped.

This middleware was developed the old days to speed up delivery of Snippets.
Given that today we have a number of different caching layers we will try
removing and see if things continue to work nicely.


glogiotatidis:
Link to error for reference https://sentry.prod.mozaws.net/operations/snippets/issues/4586320/

glogiotatidis:
Figured out that this is due to how fetchsnippetsmiddleware handles `resolve`. will fix with another approach.

PR #690, opened 2018-09-13
remove tokyo from deployments

metadave:
https://github.com/mozmeao/infra/issues/924


metadave:
https://github.com/mozmeao/snippets-stats-proxy/pull/15

PR #692, opened 2018-09-18
Add isDefaultBrowser JEXL Targeting 

glogiotatidis:
- Adds isDefaultBrowser JEXL Targeting
- Small refactor of fields to admin form and model fields.

The idea is that I'll create custom generic form fields that will output JEXL.  On form save the final JEXL expression will be calculated by joining on JEXL fields expressions. The expression will be added to AS payload as is. 

We also save individual JEXL field values in `Target.jexl` `DynamicField` to allow form initial value setting.

### DIFF
--- a/snippets/base/forms.py
+++ b/snippets/base/forms.py
@@ -568,7 +568,7 @@ class TargetAdminForm(forms.ModelForm):
         if self.instance and self.instance.jexl:
             for name, field in self.fields.items():
                 if name.startswith('filtr_'):
-                    field.initial = self.instance.jexl.get(name[6:])
+                    field.initial = self.instance.jexl.get(name)
 
     def save(self, *args, **kwargs):
         jexl = {}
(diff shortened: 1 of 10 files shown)


glogiotatidis:
/me wonders why I didn't skip the `DynamicFIeld` part and create add isDefaultBrowser as a model level field. 

I'd appreciate ideas / feedback on the different approaches

snippets/base/forms.py
```diff
@@ -534,3 +534,51 @@ def clean(self):
...
+    def save(self, *args, **kwargs):
+        jexl = {}
+        jexl_expr_array = []
+
+        for name, field in self.fields.items():
+            if name.startswith('filtr_'):
+                jexl[name[6:]] = self.cleaned_data[name]
+                jexl_expr_array.append(field.to_jexl(self.cleaned_data[name]))
+        self.instance.jexl = jexl
+        self.instance.jexl_expr = ' && '.join([x for x in jexl_expr_array if x])
```
pmac: I like the dynamic field approach just fine. I do think that keeping the serialization of the jexl dynamic columns to `jexl_expr` in the `save()` method on the model makes the most sense as it could be confusing if this model were updated and saved someday outside of this form. I also think adding a `spec` to the `jexl` dynamic field definition is a good thing to do with these things since it's a bit more self-documenting and will ensure data integrity.

glogiotatidis: Thanks for the input and good point on the spec, added.

glogiotatidis:
Thanks for your time @pmac

PR #704, opened 2018-09-21
[Fix #673] Add Profile Age JEXL targeting.

### LINKED ISSUE #673: JEXL Targeting: Profile Age
Waiting on ASR to provide local time in JEXL context. See https://github.com/mozmeao/snippets-service/issues/650#issuecomment-414990515 and https://bugzilla.mozilla.org/show_bug.cgi?id=1485345

glogiotatidis:


snippets/base/admin/fields.py
```diff
@@ -26,3 +29,46 @@ def to_jexl(self, value):
...
+            'minimum': '{minimum} <= {attr_name}',
+            'maximum': '{attr_name} < {maximum}'
+        }
+        self.jexl = kwargs.pop('jexl', self.jexl)
+        choices = kwargs.pop('choices')
+        fields = (
+            ChoiceField(choices=choices),
+            ChoiceField(choices=choices),
+        )
+        kwargs['widget'] = RangeWidget(choices=choices)
```
jpetto: Just checking if we should ensure the JEXL coming in from `kwargs` has all the necessary things.

Also, should we ensure `choices` exists in `kwargs`?

(It's pretty obvious that all information is present in this PR, I'm just wondering about 6 months down the road when this gets even more complex.)

glogiotatidis: Good point. I'll make choices a separate argument.

glogiotatidis:
Thanks for your time @jpetto :cake:

PR #725, opened 2018-09-26
[Fix #674] Add browser add-on JEXL targeting.

### LINKED ISSUE #674: JEXL Targeting: Addons Info
Include options
 - Installed
 - Not Installed

~~Installed and Enabled~~ See #504 
~~Installed and Disabled~~ See #504 

~~Also ping addons server API to translate addon URL to addon GUID~~ Tracked in issue #707

glogiotatidis:


snippets/base/migrations/0050_auto_20180925_1131.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0049_auto_20180925_1059'),
+    ]
+
+    operations = [
```
pmac: Any idea why this migration is here? I don't see the corresponding change to the model in this PR.

glogiotatidis: Good question: I changes the repr values of `300` and `400` in another commit and forgot to trigger a migration (b/c why does this need a db migration?). So this is valid and it doesn't really belong to this PR but I had to sneak it in.

glogiotatidis: Removed the migration from this PR to keep it clean and tidy

pmac: Ah. Cool. As long as it's here on purpose it's fine with me :) It is odd that small changes like this that aren't reflected in the DB schema need migrations...

PR #738, opened 2018-10-08
Include number of snippets in ASR payload.

glogiotatidis:


glogiotatidis:
Minor change, self merging

Issue #712, opened 2018-09-24
Make snippets-admin instances to always redirect to https://snippets-admin.mozilla.org/


glogiotatidis:
- [x] Deploy code
- [x] Add snippets-admin in allowed OIDC callbacks 
- [x] Configure snippets-admin to always redirect to snippets-admin.mozilla.org

PR #743, opened 2018-10-09
[Issue #712] Add django-enforce-host.

glogiotatidis:


glogiotatidis:
I see that for kitsune you have a custom middleware on top of `enforce_host` to handle IP addresses. Is that required?

glogiotatidis:
pmac told me that we need to allow IPs to go through for the healthchecks. Marking WIP

snippets/settings.py
```diff
@@ -23,6 +23,8 @@
 ALLOWED_HOSTS = config('ALLOWED_HOSTS', cast=Csv())
 ALLOWED_CIDR_NETS = config('ALLOWED_CIDR_NETS', default='', cast=Csv())
 
+ENFORCE_HOST = config('ENFORCE_HOST', default=None)
```
pmac: I like to use `cast=Csv()` for this one (like `ALLOWED_CIDR_NETS` above) since you can have a list of enforced hosts, but it's not a requirement.

jpetto:
Was just looking at this, too. Agree it looks good, and @pmac's suggestion makes sense. 😀

glogiotatidis:
Thanks both! :beers:

PR #715, opened 2018-09-24
[Fix #705]  Remove Template Widget Preview button for ASR Snippets

glogiotatidis:

PR #772, opened 2018-10-18
 [Issue #715] Add Slack notifications for Snippet 

glogiotatidis:


snippets/base/admin/legacy.py
```diff
@@ -162,6 +162,7 @@ def get_changelist_form(self, request, **kwargs):
         return form
 
     def save_model(self, request, obj, form, change):
+        obj.creator = request.user
```
pmac: Does this mean that the "creator" will be whoever last saved it? I think you'd need to do:
```suggestion
        if not obj.creator:
            obj.creator = request.user
```

snippets/base/slack.py
```diff
@@ -0,0 +1,24 @@
...
+                            context={'snippet': snippet})
+    _send_slack(data)
+
+
+def _send_slack(data):
+    if not (settings.SLACK_ENABLE and settings.SLACK_WEBHOOK):
+        logger.info('Slack is not enabled.')
+        return
+
+    response = requests.post(settings.SLACK_WEBHOOK, data=data,
```
pmac: Note: you could instead just build a `dict` and use `requests.post(url, json=data)` and it will `json.dumps` for you and set the proper content type. This works fine though if you just like using Jinja better.

snippets/base/admin/legacy.py
glogiotatidis: of course, great catch!

For `Snippets` we allow `creator` to be null, since we have all these snippets what have been created before we added the field (the new `ASRSnippets` require `creator`). So i changed the code to 

`if not obj.creator and not obj.id`

to find out to my surprise that a new obj already has an `id` by the time it reaches that point, which is before we call `super().save_model()`.

I changed the code to `if not obj.creator and not form.initial` to identify the new objects but I'm curious about your thoughts on this @pmac

snippets/base/slack.py
glogiotatidis: Aha didn't know about `json=`. I'll keep Jinja for the template functionality for this one but the requests trick is nice, thanks!

snippets/base/admin/legacy.py
pmac: There is that `change` parameter that is passed to the function that will be `False` if this form is creating a new object and `True` if it's changing an existing one. I'd think you can rely on that, but I'm not sure of the nuance. I suggested doing the more simple thing so that you could give a creator to any snippet that didn't have one already, but if that's not needed  you can probably just change it to:

```python
if not change:
    obj.creator = request.user
```

glogiotatidis: I'll take that, thanks @pmac!

glogiotatidis: I'll take that, thanks @pmac!

glogiotatidis:
Thanks!

PR #774, opened 2018-10-19
Only set creator once, when obj.creator==None.

glogiotatidis:


glogiotatidis:
Thanks @pmac, good catch!

PR #776, opened 2018-10-22
[Fix #707] Automatically fetch guid and name for Addons.

### LINKED ISSUE #707: Create automation to convert add-on URLs to add-on GUIDs
We can use the add-ons server's API to do so

 - https://addons-server.readthedocs.io/en/2018.05.17/topics/api/addons.html#detail

glogiotatidis:


snippets/base/forms.py
```diff
@@ -723,3 +725,35 @@ def save(self, *args, **kwargs):
...
+class AddonAdminForm(forms.ModelForm):
+    def clean(self, *args, **kwargs):
+        cleaned_data = super().clean(*args, **kwargs)
+        if 'url' in self.changed_data:
+
+            # Add ending slash and remove query params when needed.
+            url = cleaned_data.pop('url')
+            if not url.endswith('/'):
+                url += '/'
+            url = url[:-(len(urlparse(url).query) + 1)]
```
jpetto: Should these steps be reversed? Currently, if you have a URL with a querystring, a trailing slash gets added after the querystring (e.g. `?foo=bar&boom=bap/`), then the query string _and_ the trailing slash get stripped by the `urlparse` line.

snippets/base/models.py
```diff
@@ -660,11 +661,35 @@ class Addon(models.Model):
...
     class Meta:
         ordering = ('name', )
 
     def __str__(self):
         return self.name
+
+    @staticmethod
+    def fetch_details(url):
+        if not url.endswith('/'):
+            url += '/'
```
jpetto: I guess it doesn't hurt to re-check for a trailing slash here, but should we be performing this logic both here and in the `clean` method in `forms.py`?

snippets/base/forms.py
glogiotatidis: Great catch! I updated the code to handle this correctly and also removed the ending slash check from the other method

glogiotatidis:
Updated code and added a few more tests. Thanks @jpetto !

glogiotatidis:
Hold your re-review @jpetto . refining the code a bit. thanks!

Issue #767, opened 2018-10-17
Fix CSRF timeout


glogiotatidis:
When clicking to save a Snippet sometimes a CSRF exception is returned. This happens in production.

We have session CSRF ([code](https://github.com/mozmeao/snippets-service/blob/master/snippets/settings.py#L208))

and we use Redis to store our sessions

glogiotatidis:
@mozmeao/backend any ideas why this might be happening. I haven't found a way to reproduce this consistently.

glogiotatidis:
> Is this happening to people using the admin, or are you only seeing this in Sentry?

This is not raising a Sentry error, i"ve seen it when trying to save in admin

glogiotatidis:
@metadave @jgmize do you have any ideas?

pmac:
sessions timing out? for how long is the session valid?

jgmize:
I think we should consider switching `CSRF_USE_SESSIONS` to the default of `False` and explicitly set `CSRF_COOKIE_SECURE` to `True`.

jgmize:
We can also leverage the support for same-site cookies that landed in [Django 2.1.2](https://code.djangoproject.com/ticket/27863) and [Firefox 60](https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/)

jgmize:
Filed https://github.com/mozmeao/snippets-service/pull/784 with the changes I described above.

glogiotatidis:
Pushed change to production and pinged LCM team to keep an eye for new errors. Will close in two weeks time if no new errors.

Thanks for your help @mozmeao/theconn :metal: 

/remind me to re-eval the situation in two weeks

glogiotatidis:
While using Production Admin, configured to use CSRF cookies, I got the error again.

Any ideas @mozmeao/theconn ?

jgmize:
@glogiotatidis can you share any more details about the CSRF error you got? I double checked the settings and they looked correct, and sentry and didn't show any recent CSRF errors, and without a traceback or even an error message building a model of what could be going wrong is challenging.

jgmize:
from https://sentry.prod.mozaws.net/operations/snippets/issues/4921873/?query=is:unresolved

Forbidden (CSRF cookie not set.): /admin/base/asrsnippet/6/change/

{"status_code":403,"request":"<WSGIRequest: POST '/admin/base/asrsnippet/6/change/'>"}

[...]

PR #784, opened 2018-10-24
Use secure same-site cookies for CSRF by default for #767

jgmize:


jgmize:
The failing check on this PR is odd, as it's for a vulnerability in requests 2.18.4 but we require requests 2.19.1 in https://github.com/mozmeao/snippets-service/blob/master/requirements.txt#L152

glogiotatidis:
Interesting. I remember Session based CSRF being a security suggestion b/c we're using `mozilla.org` subdomains and cookies could be misused to cause trouble. Is this solved by the `same site` directive?

jgmize:
> Interesting. I remember Session based CSRF being a security suggestion b/c we're using mozilla.org subdomains and cookies could be misused to cause trouble. Is this solved by the same site directive?

That's exactly right, and although `same-site` is not supported in older browsers, I believe we can safely expect Mozilla employees using the snippets admin to be running at least the latest release of Firefox, if not Nightly. For more details see https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/.

glogiotatidis:
OK, let's try this out. Thanks! :cake:

PR #785, opened 2018-10-25
Set snippet.creator in form.save() to have full context on Slack.

glogiotatidis:
Slack messages didn't `snippet.creator` because it was being set in
`SnippetAdmin.save_model` which is called after `SnippetAdminForm.save`.

Setting snippet.creator in `SnippetAdminForm.save` solves this issue.

### DIFF
--- a/snippets/base/admin/legacy.py
+++ b/snippets/base/admin/legacy.py
@@ -162,9 +162,6 @@ class SnippetAdmin(QuickEditAdmin, BaseSnippetAdmin):
         return form
 
     def save_model(self, request, obj, form, change):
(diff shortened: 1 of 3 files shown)


glogiotatidis:
@pmac simple change. flagging you for review in case you have better ideas :)

glogiotatidis:
I'm going to merge this since it's really simple. Always interested at your comments pmac

pmac:
Hey. Sorry I failed to get back to you on this one. It looks pretty straight-forward to me. I like moving that logic from the admin to the form. 👍

PR #819, opened 2018-10-31
[Fix #793] Open publish_{start,end} help_text links in new tab.

### LINKED ISSUE #793: Set `target="_blank"` for time.is link in admin
The link "See the current time in UTC" on the Change ASR Snippet page should be target="_blank" (because otherwise, when you click the link, unsaved work is lost when the url loads)

glogiotatidis:


glogiotatidis:
Minor change, self merging

PR #849, opened 2018-11-19
[Fix #844] Add filter to distinguish between AS and non AS snippets.

### LINKED ISSUE #844: Add filter to distinguish between AS and non AS snippets.

glogiotatidis:


jpetto:
Looks like there's a migrations error to fix?

PR #854, opened 2018-11-21
Issue 688 create migration script

glogiotatidis:


glogiotatidis:
Apologies for this huge PR :angel:

glogiotatidis:
Note that this code will practically a one-off and will be removed in 2019q1

snippets/base/admin/actions.py
```diff
@@ -1,8 +1,216 @@
...
+def migrate_snippets_action(modeladmin, request, queryset):
+    for snippet in queryset:
+        if not snippet.on_startpage_5:
+            messages.error(request, 'Only Activity Stream Snippets can be migrated.')
+            return
+
+    migrated_snippets = 0
+    skipped = 0
+    try:
+        with transaction.atomic():
```
pmac: you've already decorated this function with `transaction.atomic`. redundant?

snippets/base/admin/actions.py
```diff
@@ -1,8 +1,216 @@
...
+
+    form = TargetAdminForm(instance=target, data=data)
+    form.is_valid()
+    target = form.save()
+
+    asrsnippet.target = target
+
+
+def _migrate_locales(snippet, asrsnippet):
+    for locale in snippet.locales.all():
```
pmac: could just do `asrsnippet.locales.add(*snippet.locales.all(), bulk=True)` to be a bit quicker, but it's not a big deal or blocker. Mostly style I think.

snippets/base/admin/fields.py
```diff
@@ -118,7 +116,9 @@ class JEXLFirefoxRangeField(JEXLRangeField):
     def __init__(self, **kwargs):
         # Include only versions greater than 63, where ASRSnippets exist.
         min_version = 64
-        max_version = int(product_details.firefox_versions['FIREFOX_NIGHTLY'].split('.', 1)[0])
```
pmac: What makes this method not sufficient?

snippets/base/migrations/0062_set_asrsnippet_id_autoincrement_starting_point.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0061_auto_20181116_0810'),
+    ]
+
+    operations = [
+        migrations.RunSQL(['ALTER TABLE base_asrsnippet AUTO_INCREMENT=10500;'], [''])
```
pmac: Why is this necessary?

snippets/settings.py
```diff
@@ -363,3 +363,5 @@
         },
     },
 }
+
+DATA_UPLOAD_MAX_NUMBER_FIELDS = config('DATA_UPLOAD_MAX_NUMBER_FIELDS', default=1000, cast=int)
```
pmac: I don't see this setting referenced anywhere else in this diff.

glogiotatidis: I have a comment in the commit about it but I also added one here. tl;dr when migrating the hundreds of snippets django raises a SuspiciousOperationError. We need to increase this number to make this work

snippets/base/migrations/0062_set_asrsnippet_id_autoincrement_starting_point.py
glogiotatidis: I want Snippets and ASRSnippets to have the same IDs for metrics. To avoid ID conflicts with new ASRSnippets I move the auto-increment point beyond the current Snippets point (which is around 9800 now)

snippets/base/admin/fields.py
glogiotatidis: Well, I have some trouble getting this to work nicely on production and/or tests. It should work in theory but in practice it requires that we spend time that we don't really have now, for a little gain. I have #855 to track fixing this in the future.

snippets/base/admin/actions.py
glogiotatidis: I like that. You get :100: style points!

snippets/base/admin/actions.py
glogiotatidis: Looks like it but it's not really. That's how the Django docs are suggesting to do it, to be able to handle the exceptions nicely

https://docs.djangoproject.com/en/2.1/topics/db/transactions/#controlling-transactions-explicitly

snippets/base/admin/actions.py
glogiotatidis: hmm that doesn't seem work. i'll leave it as is for now.

snippets/base/migrations/0062_set_asrsnippet_id_autoincrement_starting_point.py
pmac: Ah! Makes sense. This ensures no conflicts with the ported snippets. 👍

snippets/base/admin/fields.py
pmac: Cool. In that new PR don't forget about the `Version` class in the `product_details` lib.

```python
from product_details.version_compare import Version
from product_details import product_details as pd

max_version = Version(pd.firefox_versions['FIREFOX_NIGHTLY']).major
```

That's the equivalent of what you had but should be more robust. Something to try for later. I agree that doing this simple thing for now that gets this done is good 👍

PR #857, opened 2018-11-23
[Fix #856] Don't include empty data fields on ASRSnippets render.

### LINKED ISSUE #856: Don't include empty data fields on ASRSnippets render

glogiotatidis:


glogiotatidis:
Oneliner, tests pass, self merging

PR #886, opened 2018-12-11
[fix #875] Support multiple targets for each ASRSnippet.

### LINKED ISSUE #875: Allow multiple targets be to selected for one ASRSnippet

glogiotatidis:


glogiotatidis:
> Code looks good to me, but a fair number of failing tests due to references to removed `target`.

Well, I guess that happens if you don't push the lastest commit, right? :man_facepalming: 

It should be fine now

PR #897, opened 2019-01-14
[Fix #896] Remove migration code.

### LINKED ISSUE #896: Remove migration code
Remove the migration code introduced in #688

glogiotatidis:
Migration is complete, no need to keep around that code


glogiotatidis:
Thanks!

PR #901, opened 2019-01-15
[Fix #838][Fix #771] Send Slack notification on ASR Snippet status changes.

### LINKED ISSUE #771: Integrate Slack to post updates when ASRSnippet changes
See also #751

### LINKED ISSUE #838: Slack notifications for changes in the list view + include creators name
Related to #751  (let me know if you rather I reopen that issue):

1) We noticed that when we're in the list view page and checkmark the "published" box and then save, it does not send notifications to our Slack channel #snippets-admin . We would like this method to also send notifications. It only sends notifications when we're in the editing mode page and selecting "published" or "ready for review" in the drop down menu and then saving. 

2) Is it possible within the slack notifications for #snippets-admin to include the content creator's name? For example "Snippet #9553 is ready for review!" by Michelle Anderson, or "Snippet #9553 is published!" by Jean Collings. 

3) Are people who have rights to the snippet service (whether they are admin, advanced user, content creator) automatically added to the #snippets-admin channel? Is there a way to do that?

glogiotatidis:


snippets/base/admin/adminmodels.py
```diff
@@ -261,9 +260,25 @@ def get_form(self, request, obj=None, **kwargs):
...
     def make_published(self, request, queryset):
-        for snippet in queryset:
-            snippet.status = STATUS_CHOICES['Published']
-            snippet.save()
+        clean_queryset = queryset.exclude(status=models.STATUS_CHOICES['Published'])
+        no_snippets = clean_queryset.count()
+        no_already_published_snippets = queryset.count() - no_snippets
+
+        snippets = []
+        with transaction.atomic():
```
jgmize: This makes the `transaction.atomic` decorator on the method redundant (and we wouldn't want the DB to get locked waiting for responses from slack), so I assume you just forgot to remove it?

glogiotatidis: Yes i forgot to remove it here, good catch. It's worth mentioning though that if I was to handle any possible exceptions, this is the way to do it according to the [docs](https://docs.djangoproject.com/en/2.1/topics/db/transactions/) as you shouldn't be handling exception within the atomic block.

PR #903, opened 2019-01-17
[Fix #804] Add ASRSnippet iCal feed with filtering + extra admin field

### LINKED ISSUE #804: Update iCal / Google Calendar feed to show only certain snippets
Per #649, the iCal / Google calendar integration is working. However, this shows all snippets running which includes some snippets that we will always have on [does not have start/end date] and we want to be able to filter those out. 

So question:
1) Can we just get the feed that shows snippets that have a start and end date?
2) OR use what we currently have but be able to filter ourselves that we only want to see snippets that have a start and end date. [I have a feeling that this is a no but wanted to ask anyways].

glogiotatidis:


snippets/base/feed.py
```diff
@@ -1,25 +1,78 @@
...
     def item_title(self, item):
         return item.name
 
+    def item_link(self, item):
+        return item.get_admin_url()
+
     def item_description(self, item):
-        return item.name
+        description = '''
+Channels: {}
```
pmac: Just a note: I like to use the [textwrap.dedent](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) function on this, just for code asthetics.

snippets/base/urls.py
```diff
@@ -21,5 +21,5 @@
     path('csp-violation-capture', views.csp_violation_capture, name='csp-violation-capture'),
     path('healthz/', watchman_views.ping, name='watchman.ping'),
     path('readiness/', watchman_views.status, name='watchman.status'),
-    path('feeds/snippets-enabled.ics', EnabledSnippetsFeed()),
```
pmac: Is anyone using this URL, and if so will removing it break things?

glogiotatidis: Not really, this ended up not being used b/c it's wasn't filtered

glogiotatidis:
Thanks!

PR #915, opened 2019-02-11
Make Campaign.slug field required.

glogiotatidis:


glogiotatidis:
Nothing interesting here. Tests pass, self merging.

Issue #887, opened 2018-12-11
Automated data export to Google Cloud Storage for Analytics 


glogiotatidis:
We need an automated way to upload to Google Cloud Storage a file with ASRSnippet metadata. The metadata will be in turn parsed by a Python based App Engine job that loads things into Big Query. 

Open questions:
 - [x] File format: Probably CSV. We need to pick a good delimiter 
  - CSV, Delimeter `,`, all fields quoted
 - [x] Interval: How often should the file get updated
  - Daily
 - [x] Metadata to be included:
  - ID
  - Name
  - Campaign

Requires:
- [x] ~~Google Cloud Storage credentials and bucket name~~. Will use S3

- [x] DMS Monitoring Prod

- [x] ADR Document

glogiotatidis:
@gkabbz is this still a thing? Do you have deadline you want this delivered?

gkabbz:
@glogiotatidis apologies for the delay. Yes it's still a thing. I've outlined some requirements here: https://docs.google.com/document/d/1uJkQQkbNZnM-bWmhPZ4AgZPkKZ8wCJXxmD3aZdZMKik/edit
Let me know if you have any questions. I'll setup the bucket for where to drop the file as well as a service account to give you access to the bucket next week.

glogiotatidis:
Done!

PR #921, opened 2019-02-26
[Issue #887] Snippet metadata export cron job.

glogiotatidis:


glogiotatidis:
@pmac note that i'm using Google's S3 compatibility layer. What we want to do is simple and does not require any google specific functionality. We also avoid adding more python dependencies this way. Thoughts?

pmac:
I like it 👍 I assume there's a reason we can't just use S3? Either way, that seems fine.

glogiotatidis:
> I assume there's a reason we can't just use S3?

I'm in discussions with the Data team to use S3 if possible

glogiotatidis:
Waiting for final decisions on where we'll store the metadata file.

glogiotatidis:
@pmac we changed that to upload to S3 directly. 

I plan to re-use the existing buckets and configure a random url like

   `/metadata/wu1me1no4de8ri1ca3cy/snippets_metadata_20190402.csv`

to share with the data egnineers.

can you please take a second look?

docs/architecture/decisions/0002-export-asrsnippet-metadata-for-snippet-metrics-processing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Accepted
+
+## Context
+
+Data Engineers are building performance dashboards for Snippets using the
+metrics we collect from Firefox Telemetry. Telemetry pings include only basic
+information about the Snippet, like Snippet ID.
+
+For better to understand ad more complete dashboards, we want to enhance the
```
metadave: can you fix this sentence?

        For better to understand ad more complete dashboards

glogiotatidis: done, thanks!

snippets/base/storage.py
```diff
@@ -10,8 +10,17 @@
 
 
 class OverwriteStorage(FileSystemStorage):
+    """
+    Comes from http://www.djangosnippets.org/snippets/976/
+    (even if it already exists in S3Storage for ages)
```
pmac: I'm confused about the mention of S3Storage here. This `OverwriteStorage` class doesn't seem to have anything to do with the S3 storage stuff.

glogiotatidis: Fair comment. Not sure why this is there. I just refactored / copy-pasted from `snippets.storage` to `snippets.base.storage`. Will remove this comment make things clear.

glogiotatidis:
Thanks @pmac!

Issue #940, opened 2019-04-01
Remove dependency on DynamicField


glogiotatidis:
The project has a dependency on [MariaDB's DynamicField](https://django-mysql.readthedocs.io/en/latest/model_fields/dynamic_field.html).

Evaluate its replacement, probably with a simple django model fields, to remove the MariaDB dependency.\

Used in 
 - https://github.com/mozmeao/snippets-service/blob/master/snippets/base/models.py#L569
 - https://github.com/mozmeao/snippets-service/blob/master/snippets/base/models.py#L288

Migration Plan:
 - [x] Set `ENABLE_ADMIN=False` on `snippets-admin`
 - [x] Backup
 - [x] Capture the responses for `en-US`, `de`, `fr` on `release` for both `ASR` and `AS` snippets
 - [x] Merge and push all the way to production #950:
    - Generates all the new fields and migrates data to the new fields.
    - Does not remove anything yet
 - [x] Validate the responses for `en-US`, `de`, `fr` on `release` for both `ASR` and `AS` snippets
 - [x] Backup again
 - [x] Merge and push all the way to production #955 
 - [x] Merge and push all the way to production #956
 - [x] Set `ENABLE_ADMIN=True` on `snippets-admin`

glogiotatidis:
Moving off DynamicField will enable us to move off MariaDB and its unicode incompatibilities. See also #910

glogiotatidis:
We discussed and agreed with @jgmize to prioritize this work so we can unlock the options of GCP and Postgres

glogiotatidis:
This was completed successfully.

For the record:
 - I exec-ed into a running pod
 - Run `./manage.py dump > dump.json` to get all the data
 - Run `export DATABASE_URL=postgresql://...`
 - `./manage.py migrate`
 - `./manage.py shell_plus` and `ContentType.objects.all().delete()`
 - `./manage.py load dump`

Then I used `deis` to redeploy with the new `DATABASE_URL`. 

Updated on all `dev`, `stage`, `admin`, `prod` frankfurt and oregon

glogiotatidis:
Logged in / out from /admin, generated new bundles, everything appears to be working fine

PR #950, opened 2019-04-04
[Issue #940] Migrate Snippet off DynamicField.

glogiotatidis:


glogiotatidis:
Note: ` 'has_testpilot'` was intentionality not migrated b/c it has been migrated to `has_addon` in the past.

glogiotatidis:
This gets us off Dynamic Field. I still need to fine tune things to migrate to postgres. 

@jgmize if you're in a hurry I suggest that we merge this early next week and we go ahead with mariadb / mysql on GCP for now. If you can wait another week we can get postgres.

jgmize:
Thanks @glogiotatidis. I can wait another week for postgres :)

jpetto:
The code here all looks good, but I haven't tested it locally yet. Been having serious Docker issues today. Sounds like we're waiting for the postgres changes to come in anyway. 😄

glogiotatidis:
Add two new PRs (#955, #956) to make the app completely db agnostic and therefore capable to migrate to pgsql. Even tried sqlite locally and worked fine!

We need to take special care while migrating, I have a detailed migration plan here https://github.com/mozmeao/snippets-service/issues/940#issue-427538060

PR #995, opened 2019-05-09
[Fix #994] Fix NewRelic reporting.

### LINKED ISSUE #994: Fix NR reporting
See https://github.com/mozmeao/infra/issues/1106

glogiotatidis:


snippets/wsgi/app.py
```diff
@@ -6,22 +6,19 @@
 For more information on this file, see
 https://docs.djangoproject.com/en/1.7/howto/deployment/wsgi/
 """
-import os
-os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'snippets.settings')  # NOQA
-
-from django.core.wsgi import get_wsgi_application
-
 import newrelic.agent
 from decouple import config
```
jgmize: I'm not sure about the impact of importing `config` from the `decouple` module before calling `newrelic.agent.initialize` below. Instead of importing `config` or `os` to enable customization via environment variables, in bedrock we decided to eliminate that customization point in https://github.com/mozilla/bedrock/pull/7150

glogiotatidis: Cool, updated!

PR #1025, opened 2019-06-05
[Fix #1024] Add mass unpublish action.

### LINKED ISSUE #1024: Add  mass unpublish in the list view

glogiotatidis:


snippets/base/admin/adminmodels.py
```diff
@@ -619,31 +620,50 @@ def get_form(self, request, obj=None, **kwargs):
...
-                snippets.append(snippet)
+        now = datetime.utcnow()
+        # Create a list of matching snippets before altering them to log action
+        # and send Slack messages.
+        snippets = [snippet for snippet in clean_queryset]
+        clean_queryset.update(status=status, modified=now)
 
         for snippet in snippets:
-            slack.send_slack('asr_published', snippet)
+            self.log_change(request, snippet, 'Changed status.')
```
glogiotatidis: @pmac i remember us talking about logging user actions: all Django Admin actions are logged by default and one can use `self.log_{change,addition,deletion}` to log custom actions, like in this case.

Thought you might find this interesting. Other than that, this PR is mostly moving code around to add an additional snippet unpublish action, next to the existing snippet publish action.

snippets/base/admin/adminmodels.py
```diff
@@ -619,31 +620,50 @@ def get_form(self, request, obj=None, **kwargs):
...
-        snippets = []
-        with transaction.atomic():
-            for snippet in clean_queryset:
-                snippet.status = models.STATUS_CHOICES['Published']
-                snippet.save()
-                snippets.append(snippet)
+        now = datetime.utcnow()
+        # Create a list of matching snippets before altering them to log action
+        # and send Slack messages.
+        snippets = [snippet for snippet in clean_queryset]
```
pmac: NIT: same as `snippets = list(clean_queryset)` no? Both are fine I think, just a bit verbose this way in my opinion.

glogiotatidis: Sure it makes more sense

Issue #997, opened 2019-05-14
Make snippet localization easier


glogiotatidis:
Original bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=870491

Althought I don't believe we will support in app localization anytime soon, we have make things easier by setting the default values on fields like the block button #757 

Let's explore options.

glogiotatidis:
This is done. Will work on #757 to add more localized strings

PR #1045, opened 2019-06-20
[Issue #997] Migrate TargetedLocales to Locale

glogiotatidis:


snippets/base/models.py
```diff
@@ -1401,6 +1440,7 @@ class ASRSnippet(models.Model):
             'See the current time in <a target="_blank" href="http://time.is/UTC">UTC</a>'))
 
     locales = models.ManyToManyField('TargetedLocale', blank=True, verbose_name='Targeted Locales')
```
jpetto: Do we still need `locales`?

glogiotatidis: Yes, for now. I want to run a data migration script manually after this reaches production and then I'll pull request a migration to remove `locales`.

Issue #985, opened 2019-05-08
Optimize PNGs


glogiotatidis:
Make sure that the uploaded PNGs are optimized. Pillow library supports `optimize` flag.

https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html?highlight=webp#png

## Deploy plan
 - [x] Optimize existing images. 
 - [x] Push code

glogiotatidis:
PIllow's `optimize` didn't do much, but [`pngquant`](https://pngquant.org/) is amazing. 

Current library can be optimized down to 1.7Mb from 3.7Mb with no visual difference.

glogiotatidis:
Completed

PR #1047, opened 2019-06-20
[Issue #985] Optimize PNG Images.

glogiotatidis:


glogiotatidis:
Hmmm experimenting with refactoring all the checks and optimizations in `Model.clean()` to have everything in one place instead of some code in field validators and other in form clean.

Any ideas on form clean vs model clean @pmac?

glogiotatidis:
Updated

snippets/base/models.py
```diff
@@ -602,6 +607,46 @@ def snippets(self):
...
+                stdout=subprocess.PIPE
+            )
+            if cmd.stdout:
+                new_image = io.BytesIO(cmd.stdout)
+                self.image.file = InMemoryUploadedFile(
+                    new_image, 'ImageField', self.image.name,
+                    'image/png', len(new_image.read()), None
+                )
+
+        if settings.IMAGE_MAX_SIZE > 0 and self.image.size > settings.IMAGE_MAX_SIZE:
```
pmac: is this max size checking the new size or the uploaded size? I see above that optimize has updated `self.image.file`, but I'm not sure if that updates the other metadata like `self.image.size`. I'm just wondering if the size check should come before the optimization step so that work can be avoided for files that are too large.

glogiotatidis: Good catch. Actually `self.image.size` points to `self.image.file.size` in this case ([django code](https://github.com/django/django/blob/044cc544203fdcf2601ff200547ecae2e51a5db5/django/db/models/fields/files.py#L64)) so we're good. 

I want the size check after the optimization so things are simpler for the Editors. All we need is the final file size to be within limits, either pre-optimized or optimized by the django app.

glogiotatidis:
Rebased

glogiotatidis:
Thanks!

PR #1050, opened 2019-06-27
Update clock-deploy.yaml from bin/update-config.sh

jgmize:
This should only be merged once `clock-deploy.yaml` exists for each namespace in snippets-config.


jgmize:
it turns out the clock procs have been scaled to 0 in production, so this may not be needed. @glogiotatidis are the clock procs needed at all at this point?

glogiotatidis:
> it turns out the clock procs have been scaled to 0 in production, so this may not be needed. @glogiotatidis are the clock procs needed at all at this point?

There's running clock in `snippets-admin` namespace. Definitely needed

jpetto:
mozmeao/infra#986

PR #1054, opened 2019-06-28
Add docs for git push triggered deployments

jgmize:


jgmize:
@glogiotatidis please note that the `.gitlab-ci.yml` currently uses `stage` and `prod` branch names for the deployments, which is consistent with other projects but different than the previous Jenkins CI/CD pipeline, but I left that out of these docs, electing to link to the file instead. Let me know if you think that's sufficient or we should make that mapping explicit in the docs. If you have a preference for the previous spelling (or want to accept either spelling, or want to use something completely different, etc) then it's fairly straightforward to modify the yaml to support other branch names.

jgmize:
This PR was obsoleted by https://github.com/mozmeao/snippets-config/pull/28

PR #1061, opened 2019-07-03
Update additional config branches for more targeted deployments

jgmize:
Together with https://github.com/mozmeao/snippets-config/pull/26 this gives us an automatically updated master branch while only triggering one deployment at a time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,8 +12,7 @@ dev:
     - bin/flake8.sh
     - bin/coverage-tests.sh
     - bin/push2dockerhub.sh
-    - CLUSTER_NAME=oregon-b bin/update-config.sh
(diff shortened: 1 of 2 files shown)


glogiotatidis:
This looks good. I want to make myself comfortable with the new ci and possibly fine  tune a few more bits. If you don't mind let's hold changes for a day or two until I wrap my head around this.

jpetto:
@pmac Can you give this (and the related PR) a look over in terms of approach? I think it'd be good to get the team on board with changes from this point forward.

pmac:
Very cool Giorgos!

jgmize:
thanks @glogiotatidis, https://github.com/mozmeao/snippets-config/pull/28 is a much better approach.

PR #1099, opened 2019-07-25
 [Fix #1067] Schedule snippets in the future to allow undo. 

### LINKED ISSUE #1067: Schedule snippets at least 5 minutes ahead of now() to allow undo
When snippets are set to publish now, automatically set the publication time to `now()+5'` to allow Editors to undo publication if they spot an error.

glogiotatidis:
Merge after #1093

### DIFF
--- a/snippets/base/tests/test_models.py
+++ b/snippets/base/tests/test_models.py
@@ -667,15 +667,40 @@ class JobTests(TestCase):
         self.assertTrue(job.channels, set(['release', 'beta', 'nightly']))
 
     def test_clean(self):
-        job_clean = JobFactory.create(publish_start=datetime.utcnow() - timedelta(days=1),
-                                      publish_end=datetime.utcnow())
+        job_clean = JobFactory.create(publish_start=datetime.utcnow() + timedelta(days=1),
+                                      publish_end=datetime.utcnow() + timedelta(days=2))
         job_clean.clean()
 
-        job_dirty = JobFactory.create(publish_start=datetime.utcnow(),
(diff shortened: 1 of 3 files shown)


glogiotatidis:
@pmac rebased and ready for review. 

I want to create a publishing "undo" functionality by scheduling the job, X minutes in the future. If an Editor publishes something by mistake then they have time to fix it. Thoughts?

snippets/base/tests/test_models.py
```diff
@@ -667,16 +667,40 @@ def test_channels(self):
...
 
+    @override_settings(SNIPPETS_PUBLICATION_OFFSET=5)
+    def test_clean_publication_offset(self):
+        utcnow = datetime.utcnow()
+
+        job_no_publish_start = JobFactory.create(publish_start=None, publish_end=None)
+        with patch('snippets.base.models.datetime') as datetime_mock:
+            datetime_mock.utcnow.return_value = utcnow
+            job_no_publish_start.clean()
+        job_no_publish_start.publish_start = utcnow + timedelta(5)
```
pmac: These look like assignments but I think they should be assertions. Also I think you need to use the kwarg: `timedelta(minutes=5)`.

glogiotatidis: Ermm of course. Thanks!

glogiotatidis:
> Email programs (gmail e.g.) have had this kind of feature for a while and it seems like an excellent concept for snippets. Somehow hitting that "submit" or "publish" button always makes me realize that I've made some mistake, but not before :)

haha exactly what happens with me too :) Thanks for your time pmac!

Issue #1005, opened 2019-05-23
Update gunicorn setup 


glogiotatidis:
Explore 

 - ideas found in "[Docker is different](https://pythonspeed.com/articles/gunicorn-in-docker/)" blog post. 
 - Meinheld or other workers.

glogiotatidis:
Interested to try gunicorn + eventlet and gunicorn + gthread in October.

Want to drop Meinheld b/c it's not clear if and how the project is maintained. 

Load of snippets-service will be mostly from Django Admin usage after #917 

/remind me in October

glogiotatidis:
/remind me to check in newrelic apm before and after temp dir change in production  on September

glogiotatidis:
No significant changes in apm related to this change.

PR #1111, opened 2019-08-05
[Issue #1005] Set Gunicorn to use /dev/shm for temp dir.

glogiotatidis:


metadave:
are you expecting a performance change here that we can measure?

glogiotatidis:
> are you expecting a performance change here that we can measure?

probably not but i'll take a note to check newrelic before and after

/remind me on September

Issue #1110, opened 2019-08-02
Add sign up to Monitor special action 


glogiotatidis:
Related ASR PR https://github.com/mozilla/activity-stream/pull/5220/files/5542e7424581c54cf5ca3a60477fa385fab81021

See also #1102 

## Check list
 - [x] Confirm with FxA team that flow metrics used are OK

jgmize:
After deploying #1103 to production, @jkt230 and I created a test snippet with a button and the [URL from the example docs](TODO). Clicking on the button took me to the URL, but when I tried to sign in I got the following error:

![image](https://user-images.githubusercontent.com/161718/63372125-1db6e600-c34b-11e9-8202-b2c546447e15.png)

jgmize:
To resolve the issue above we needed to change the dynamic `utm_content=[[channels]]` part of the url to a temporarily hard coded `utm_content=nightly`. The targeting section of the admin seems to have gone missing as part of the deployed changes.

jgmize:
Jean & I figured out that the targeting section of the admin got moved to the Jobs section, so this is a non-issue for snippets that get published by Jobs. I granted Job permissions to the `ASR Snippets` group so that Jean and other ASR Snippets members can view & add jobs.

jkt230:
Actually tried signing in with an email address on a snippet that has a Job published with nightly targeting which left the utm_content=[[channels]] and still received this error:
![Screen Shot 2019-08-23 at 2 41 28 PM](https://user-images.githubusercontent.com/6807556/63625355-383fc800-c5b4-11e9-9d36-cb5cd226771f.png)

glogiotatidis:
I cannot say how wrong is that you pushed to production without first consulting me or reading the issues of the code pushed to production. #1034 had a number of things to check before reaching prod and a number of migration steps that haven't happened.  I'm sure that this happened with the best intentions but still not ok.

Here you're trying to test an open issue without any related code merged. I'm not surprised that it doesn't work. 

ASR didn't have this merged in firefox before I got on PTO and thus I set a reminder to fix this by the time 71 hits Beta.

PR #1126, opened 2019-08-30
[Issue #1110] Adds ENABLE_FIREFOX_MONITOR special action.

glogiotatidis:
Adds the `ENABLE_FIREFOX_MONITOR` special action.

Also moves a few things around to enable special actions for buttons, we only had this for links before.


glogiotatidis:
@pmac i tag you for review mostly for the django / python side of things. The Activity Stream part has been tested, no need to worry about that.

snippets/base/fields.py
```diff
@@ -10,3 +10,7 @@ def __init__(self, *args, **kwargs):
                   'validators': [validate_regex]}
         myargs.update(kwargs)
         return super(RegexField, self).__init__(*args, **myargs)
+
+
+class URLField(models.CharField):
```
pmac: Why can't you use Django's built-in `URLField`?

snippets/base/models.py
```diff
@@ -668,10 +668,46 @@ def subtemplate(self):
...
+                        }
+                    }
+                    to_delete.append(key)
+
+        for key in to_delete:
+            data.pop(key)
+
+        data.update(to_append)
+
+        return data
```
pmac: There's not really a need to return `data` since you're modifying it in place. I'd suggest just for clarity that you either copy `data` at the start of this function and return the new modified object, or just call this on `data` below and rely on it being modified in place. Not required as this also works, but returning and assigning to `data` below on line 709 is really a noop.

snippets/base/fields.py
pmac: Oh, I missed that it was your custom validator. I guess there's no real advantage to having this be a subclass of Django's `URLField`. 👍

glogiotatidis: You're correct!

snippets/base/models.py
glogiotatidis: Fair point. I like to return data from my functions, so I'll go ahead and create a local copy  as you suggest. Thanks for the better code recommendation!

PR #1135, opened 2019-09-03
[fix bug 1546165] Add optional heading to snippets:

jpetto:
- FxaSignupTemplate
- NewsletterTemplate
- SendToDeviceTemplate

Depends on mozilla/activity-stream#5294.

Marking as draft until PR referenced above is approved and merged.


snippets/base/models.py
```diff
@@ -851,12 +851,23 @@ class SimpleTemplate(Template):
         on_delete=models.SET_NULL,
         verbose_name='Section Title Icon',
         related_name='simple_section_icons',
-        help_text=('Section title icon. 32x32px. PNG. '
+        help_text=('Section title icon. 16x16px. PNG. '
+                   'section_title_text must also be specified to display.'),
+    )
+    section_title_icon_dark_theme = models.ForeignKey(
```
glogiotatidis: We won't be using dark icons anytime soon according to the snippets editors, so I suggest that we don't include these options to keep it simple(r)

snippets/base/models.py
```diff
@@ -1099,6 +1110,39 @@ class FxASignupTemplate(Template):
         help_text=('The background color of the button. Valid CSS color. '
                    'Defaults to Firefox Theme Color.'),
     )
+    scene1_section_title_icon = models.ForeignKey(
+        Icon,
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
```
glogiotatidis: Let's change this to `models.PROTECT` to prevent users from accidentally removing icons from published snippets. If you can please change this on the simple snippet as well.

jpetto:
Thanks for the review @glogiotatidis! Just added a commit that I think addresses everything.

snippets/base/models.py
```diff
@@ -848,15 +848,16 @@ class SimpleTemplate(Template):
         Icon,
         blank=True,
         null=True,
-        on_delete=models.SET_NULL,
+        on_delete=models.PROTECT,
         verbose_name='Section Title Icon',
         related_name='simple_section_icons',
-        help_text=('Section title icon. 32x32px. PNG. '
+        help_text=('Section title icon. 16x16px. PNG. '
```
glogiotatidis: Did this change?

glogiotatidis:
Maybe that now needs a db migration rebase before merging

snippets/base/models.py
jpetto: No - it's always been 16x16 on the AS side. I figured maybe just a typo on this end?

jpetto:
The help text for the 32x32/16x16 change is in the migration in this PR - unless you're referring to something else needing a db migration rebase?

jpetto:
In regards to merging, I'll let you handle the timing. Not sure how soon the AS team will do a full review/merge on that PR...

jpetto:
Ah! I see another migration came in outside of this PR. Just rebased and re-creating this migration.

glogiotatidis:
> Ah! I see another migration came in outside of this PR. Just rebased and re-creating this migration.

perfect thanks!

glogiotatidis:
> Ah! I see another migration came in outside of this PR. Just rebased and re-creating this migration.

perfect thanks!

glogiotatidis:
This is good to go. I will merge and push to production next week :taco:

Issue #917, opened 2019-02-18
Pre-generate snippet bundles


glogiotatidis:
Updated on Aug 5th with new implementation notes

## Summary

Pre-generate Snippet bundles based on 
 - Locale
 - Channel
 - Version

`Version` is going to be used for future Normandy experiments and replace the extensive (but underutilized Client Match Rules). All `Jobs` will target `default` Version by default. When an experiment needs to run, we will alter `Jobs` to target another version and push another version string to the experiment cohort via Normandy, similar to what we do now.  

Snippets request URL will change from

`https://snippets.cdn.mozilla.net/%STARTPAGE_VERSION%/%NAME%/%VERSION%/%APPBUILDID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/`

to 

`https://snippets.cdn.mozilla.net/%STARTPAGE_VERSION%/%NAME%/%LOCALE%/%CHANNEL%/default.json`

and `STARTPAGE_VERSION` will get bumped to `7`

For example

`https://snippets.cdn.mozilla.net/7/Firefox/en-us/beta/default.json`

## What is `default.json`?

Will be used when we run experiments via Normandy. Jobs will be able to target other than default.  See #1143

### Request

Because custom Firefox builds alter the `%CHANNEL%` variable we cannot directly point to a single file. Instead we will deploy Lambda@Edge functions which will convert the request URL to an expected format, which is practically change `%CHANNEL%-*` to `%CHANNEL%`. 

~~The Lambda function will run when the CDN needs to contact the Origin only, thus limiting the number of triggers.  Lambda function creation and deployment is tracked in #1112.~~

Will create a view in the Django app to redirect to the correct bundle. See #1112

### 404

For the cases that we don't have a snippet bundle for the given locale, channel combination we will return an empty json obj. This Lambda function will run after the Origin response and if the response is 404. Tracked in #1113.

## Other implementation notes
 - [x] Create a management command to regenerate all bundles
 - [x] Need a way to automatically force bundle regeneration when bundle related code changes, like the render functions. 
 
## Open Questions
 - ~~ Confirm that we can change Snippets request URL for all Quantum versions via Normandy~~ Moved to #1112 
[...]

PR #1144, opened 2019-09-09
[Fix #1143] Add Distributions and Distribution Bundles.

### LINKED ISSUE #1143: Add Distributions and Distribution Bundles
We want to be able to create a separate bundle of Snippets to run experiments with Normandy. 

In a typical Normandy experiment we modify a part of the URL used by the browser to request snippets. For pregenerated bundles this will be last part of the url, marked `distribution`

`https://snippets.cdn.mozilla.net/%STARTPAGE_VERSION%/%NAME%/%LOCALE%/%CHANNEL%/%distribution%.json`

Normaly this defaults to `default`.

We will introduce `Distributions` and `Distribution Bundles`. A Job will have to target a Distribution (e.g. default) and the Distribution Bundle can include multiple Distributions into a Snippet Bundle.

This setup allows for a Job to target a non default distribution and also for the client to get Jobs  from multiple distributions. This is the typical behavior we need, i.e. Users get the experiment snippets *and* the default snippets.

We don't expect to use this more than a couple of times per year, the vast majority of Jobs will just default to `default`.

# Deploy Plan
1. [x] Push to `admin` and `prod`
2. [x] Create `Default` Distribution Bundle and `Default` Distribution
3. [x] Link all Snippets to `Default` Distribution
4. [x] Create migration to make `Job.distribution` non-nullable and remove `Job.distributions` deprecated field.

glogiotatidis:


glogiotatidis:
@pmac ready for review. #1143 and #917 contain info about how this is supposed to work

snippets/base/models.py
```diff
@@ -1933,16 +1941,6 @@ def analytics_export(self):
         }
         return export
 
-    def save(self, *args, **kwargs):
```
pmac: I assume this means the `distributions` field will be going away in a subsequent migration? If there used to be multiple distributions per job, why is it moving to one distribution per job?

snippets/base/models.py
```diff
@@ -2083,13 +2081,15 @@ def update_asrsnippet_modified_date(sender, instance, **kwargs):
 
     elif (isinstance(instance, Campaign) or
```
pmac: NIT: I realize you aren't changing this, but this is equivalent to: `isinstance(instance, (Campaign, Target)):`

snippets/base/tests/test_models.py
```diff
@@ -586,6 +587,9 @@ def test_get_admin_url(self):
 
     def test_modified_date_updates_when_template_updates(self):
         snippet = ASRSnippetFactory()
+        # Must refresh from db to get the actual datetime stored in db which
```
pmac: that sounds like a bug that took a while to find ಠ_ಠ

snippets/base/management/commands/generate_bundles.py
```diff
@@ -45,63 +46,70 @@ def handle(self, *args, **options):
...
+                    data = [job.render() for job in bundle_jobs]
+                    bundle_content = json.dumps({
+                        'messages': data,
+                        'metadata': {
+                            'generated_at': datetime.utcnow().isoformat(),
+                            'number_of_snippets': len(data),
+                        }
+                    })
+
+                    # Convert str to bytes.
```
pmac: Just an FYI. You may know this, but just in case you don't, [Django has some very useful functions for str/bytes stuff](https://docs.djangoproject.com/en/2.2/ref/utils/#module-django.utils.encoding) that just make it a little easier e.g. `force_bytes` and `force_string` as well as some counterparts that respect lazy objects.

snippets/base/models.py
glogiotatidis: Didn't know that, thanks!

snippets/base/tests/test_models.py
glogiotatidis: Oh yes, indeed.

snippets/base/models.py
glogiotatidis: Yes, `distributions` is going away as I note in the [notes of the issue](https://github.com/mozmeao/snippets-service/issues/1143). We move the ability of including jobs to multiple distributions up to the DistributionBundle level. A Bundle can have multiple Distributions and a Job one Distribution.

glogiotatidis:
Thanks!

PR #1148, opened 2019-09-10
[fix #862] Add retry button label

### LINKED ISSUE #862: Update newsletter and s2d templates with new error_page_button_label variable
https://bugzilla.mozilla.org/show_bug.cgi?id=1510331

- [ ]  After we push update to the snippets-server we want to update existing snippets with localized retry button text.

jpetto:
Counterpart to mozilla/activity-stream#5299


snippets/base/admin/adminmodels.py
```diff
@@ -398,7 +399,8 @@ class NewsletterTemplateInline(admin.StackedInline):
         ('Extra', {
             'fields': (
                 'block_button_text',
-                'do_not_autoblock'
+                'do_not_autoblock',
+                'retry_button_label',
```
glogiotatidis: I think this should be under `error_text` in `Section 2 Main`

snippets/base/admin/adminmodels.py
```diff
@@ -468,7 +470,8 @@ class SendToDeviceTemplateInline(admin.StackedInline):
...
```
glogiotatidis: Same, I think this should be under `error_text` in `Section 2 Main`

snippets/base/admin/adminmodels.py
```diff
@@ -335,7 +335,8 @@ class FxASignupTemplateInline(admin.StackedInline):
                 'utm_term',
                 'utm_campaign',
                 'block_button_text',
-                'do_not_autoblock'
+                'do_not_autoblock',
+                'retry_button_label',
```
glogiotatidis: I didn't expect this in FxA template, since we have no `error_text` I don't think we display any error page, thus no need to the retry button. Can you verify on the ASR side if we ever display an error page?

jpetto: Unfortunately there's a bit of an explanation here.

tl;dr - I think you're right that we can omit the `retry_button_label` for FxA snippets.

The reason I included it is because on the ASR side, all snippets that follow the scene1/scene2/form submit pattern inherit from the same base template. This base template contains the retry button, so if any inheriting template fails a form submission, the base template renders the retry button.

We do not currently display an error state for FxA snippets, as the form submits directly to `https://accounts.firefox.com/`. As long as the FxA snippet template always submits to a new URL (and not back to the snippet), we can omit the `retry_button_label`.

In my ASR PR, I added a default value of "Try again" to this label, so if the inheriting template doesn't specify text, we'll have a reasonable fallback. 

If the FxA snippets ever change and submit back to themselves (to hit an FxA API or something), we may want to revisit and add `retry_button_label` to the admin. But for now, I think we're good skipping it.

glogiotatidis:
Thanks! :fish_cake:

Issue #996, opened 2019-05-14
Target by operating system


glogiotatidis:
See also original bug report https://bugzilla.mozilla.org/show_bug.cgi?id=1376637

# Todo
 - [x] Close bugzilla bug
 - [x] Update Targets that use Windows, Mac CMR to also include JEXL target.

glogiotatidis:
Changes to AS merged, need to create the filter on the snippet service side

glogiotatidis:
Complete :fireworks:

PR #1155, opened 2019-09-20
 [fix #937] Hide country input if SMS not selected. 

### LINKED ISSUE #937: Country required for Non-SMS S2D Snippets
Hi @glogiotatidis. 

I am updating Send to Device snippets for the F100 Snippet experience. I notice the new Snippet Admin tool requires a country code for all Send to Device snippets, even when SMS is not checked off.

Right now, when left blank, the country code defaults to `US`. And when I remove it, I get an error notice. This is even when SMS is unchecked. Perhaps this requirement needs to be conditional when SMS capabilites are checked. 
 

We have two types of send to device snippets.
- SMS and Email
-Email Only

Since we don't support SMS in all locales, we allow users to email themselves a download link. The requirement of a country code, prevents us from reaching all users and would require us to create a snippet for each country

Please see the example here: 
https://snippets-admin.mozilla.org/admin/base/asrsnippet/11428/change/?_changelist_filters=q%3D_week12_%2B_de_

SnippetID: 11428
Locale: DE

See screenshots.

Screen Shot 2019-03-28 at 11 30 23 AM

Screen Shot 2019-03-28 at 11 21 23 AM

Thank you!

jpetto:
UI only approach to the problem. this might be good enough?

@glogiotatidis thoughts?

### DIFF
--- a/snippets/base/static/js/admin/sms-country-toggle.js
+++ b/snippets/base/static/js/admin/sms-country-toggle.js
@@ -1,0 +1,42 @@
+(function () {
+    function bindSMSCountry() {
+        const chk_sms = document.querySelector('.form-row.field-include_sms input');
+        const country_row = document.querySelector('.form-row.field-country');
+        const country_input = document.querySelector('.form-row.field-country input');
+
+        // make sure required fields exist
+        if (chk_sms && country_input) {
+            // get default value from input, falling back to 'us' if empty
+            const country_default = country_input.value || 'us';
+            
+            function displayCountryField() {
+                country_row.classList.add('visible');
+            }
+
+            function hideCountryField() {
+                country_row.classList.remove('visible');
+                
+                // set the value back to default if user left it empty
+                if (country_input.value === '') {
+                    country_input.value = country_default;
+                }
+            }
+
+            // on page load, determine if we should show the country field
+            if (chk_sms.checked) {
+                displayCountryField();
+            }
+
+            // toggle display of country field when SMS checkbox changes
+            chk_sms.addEventListener('change', e => {
+                if (chk_sms.checked) {
+                    displayCountryField();
+                } else {
+                    hideCountryField();
+                }
+            });
+        }
+    }
+
+    document.addEventListener('DOMContentLoaded', bindSMSCountry);
+})();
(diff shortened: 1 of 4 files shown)


glogiotatidis:
It's OK for now. Make sure to list Country after the include SMS checkbox for better UX. Thanks!

snippets/base/static/js/admin/sms-country-toggle.js
```diff
@@ -0,0 +1,41 @@
...
+            const country_default = country_input.value;
+            
+            function displayCountryField() {
+                country_row.classList.add('visible');
+            }
+
+            function hideCountryField() {
+                country_row.classList.remove('visible');
+                
+                // set the value back to default if user left it empty
```
glogiotatidis: This doesn't seem to work, field remains empty and raises an error when saving

jpetto: hm, can you paste the error you're getting? it's working as expected on my end.

glogiotatidis: I get a generic "Please correct the error bellow" with no errors displayed, b/c the field is hidden. If I check the SMS option, I see the field and it's empty and the error is "This field is required".

STR:
 1. Create a new S2D snippet
 2. Fill in all required fields
 3. Make sure to check include sms
 4. Empty country field
 5. Save
 6. Snippet does not save, correctly, due to empty country field
 7. Uncheck `include sms`
 8. Save
 9. Snippet still doesn't save, incorrectly this time b/c country field is not visible and should default to `us`.

glogiotatidis:
Remember to update the commit messages!

snippets/base/static/js/admin/sms-country-toggle.js
jpetto: hm, i believe i have this fixed. however, i cannot create a snippet, as the first `Locale` drop down has no options (and is required).

how can i populate that select box? i've run the commands in the readme (`create_locales()`) and see locale data in the db (`TargetedLocale.objects.all()`).

jpetto: i figured it out while working on #996. 😄 followed the STR above and was able to successfully save the snippet.

glogiotatidis:
lgtm, thanks! :cake:

PR #1163, opened 2019-09-30
[Fix #799] List currently Published or Scheduled Jobs on a public page.

### LINKED ISSUE #799: ASR: List of running snippets for public audience
We used to use snippets.mozilla.com to guide people to live snippets so they could see what was running and what specific snippets looked like. Can we have something like this with ASR?

glogiotatidis:


glogiotatidis:
@jpetto if you have time to do a quick pass to check if i've done anything horrifying with my limited frontend skills.

thanks!

glogiotatidis:
> this all looks great to me! only suggestion i'd make is to standardize the select boxes - for example `------` means "all" for locale, but not for "include". might be better to explicitly label the selection `All` for both.

good point, fixed it!

PR #1171, opened 2019-10-02
[Fix #981] Filter based on number of addons and addon install functio…

### LINKED ISSUE #981: Targeting by # of add-ons in their profile
Similar to what we did for bookmarks, can we do targeting based on the number of add-ons a user has?

glogiotatidis:
…nality.


glogiotatidis:
simple target addition, tests pass, selfmerging

Issue #1112, opened 2019-08-05
Create view to redirect clients to the correct pregenerated bundle


glogiotatidis:
## Why

Because custom Firefox builds alter the %CHANNEL% variable we cannot directly point to a single file. Instead we will deploy Lambda@Edge functions which will convert the request URL to an expected format, which is practically change %CHANNEL%-* to %CHANNEL%

#### List of highly used settings for %CHANNEL%
- release
- release-cck-ubuntu
- beta
- release-cck-mozillaonline
- nightly
- nrelease-cck- âmintâ
- aurora
- 641165 release-cck-yahoo
- release-cck-archlinux
- release-cck-mint
- release-cck-fedora
- release-cck-euballot

This must return the actual channel (`release`, `beta`, etc)

## Tasks
- [x] Remove deprecated Lambda functions from snippets stage cdn
- [x] What happens with redirects which 404? Maybe that generates a lot of traffic? We either implement #1113 or we check on the app level if file exists in S3 and if not we return 204 instead of 302. Or we do nothing is client is fine with 404? We handle that on the CDN level. See #1113 for comments.
- [x]  Confirm that we can change Snippets request URL for all Quantum versions via Normandy
- [x] Create `empty.json` on snippets-prod s3 bucket with Content-type and Cache-contorl headers

glogiotatidis:
Hey @metadave 

If I'm not mistaken I remember you offering to setup me up with some TF to create a Lambda @ Edge function to use with the Snippets CloudFront Distribution.

I need a python  function at the Origin Request level (https://docs.aws.amazon.com/lambda/latest/dg/lambda-edge.html) for distribution d31g0e3nurrasj.cloudfront.net (stage) and another for d228z91au11ukj.cloudfront.net (prod)

Is this something you can do?

Thanks!

metadave:
> Is this something you can do?

yes, but it looks like there isn't any TF automation for the stage CDN. I'll have to see if I can import it and go from there.

metadave:
WIP here: https://github.com/mozmeao/infra/pull/1159

glogiotatidis:
We will go ahead and implement this as a Django view for the first iteration which is better known and won't generate new costs. 

Will update snippets stage CDN to remove deprecated lambdas

pmac:
[...]

PR #1201, opened 2019-10-24
[Issue #1112] Redirect to pregen bundles only ASR clients.

glogiotatidis:


glogiotatidis:
@pmac will merge to try out pregen bundles but i'd like your retrospective review when you have the time. thanks!

snippets/base/views.py
```diff
@@ -65,6 +66,10 @@ def fetch_snippet_pregen_bundle(request, **kwargs):
         f'{settings.MEDIA_BUNDLES_PREGEN_ROOT}/{product}/{channel}/'
         f'{locale}/{distribution}.json'
     )
+
+    if not default_storage.exists(filename):
```
glogiotatidis: A test run indicates that this needs to be cached, will investigate further

snippets/base/views.py
```diff
@@ -47,13 +47,14 @@ def template_name(self):
 
 
 def fetch_snippets(request, **kwargs):
-    if settings.USE_PREGEN_BUNDLES:
+    if settings.USE_PREGEN_BUNDLES and kwargs['startpage_version'] == 6:
```
pmac: Only version 6 or `>= 6`?

glogiotatidis: Will keep '6' for now. Who knows what `7` will look like

pmac: Fair.

PR #1206, opened 2019-10-25
[Fix #1041] Complete Jobs with no metrics data for the last 24h.

### LINKED ISSUE #1041: Create Freq Capping Goals for Jobs
Create Global Frequency Capping goals for Jobs (introduced in #1034). 

Goals:
 - Total impressions
 - Total clicks and button clicks (combined)
 - Total blocks

A `Job` can have multiple goals (freq capping, datetime), first one reached will un-publish the `Job`.

Requires #1036. Also `Job`s with Freq Cap goals must get data from Redash more often (TBD).

## Use case:
 - Publish a Snippet for 1M impressions. It's fine it snippet gets 5K-10K impressions more.

## Questions
 - [x] How do we handle Jobs with Freq Capping when no Telemetry data is available? 
   - Jobs that have impression, block or click limits set and no metrics data updated for at least 24 hours will be marked Complete.

glogiotatidis:


snippets/base/tests/test_commands.py
```diff
@@ -178,34 +180,58 @@ def test_base(self):
...
             limit_clicks=100,
             metric_clicks=10,
+            metric_last_update=now - timedelta(hours=1),
             publish_end=datetime.utcnow() + timedelta(days=1))
         job_block_limit_not_reached = JobFactory(
             status=models.Job.PUBLISHED,
             limit_blocks=100,
             metric_blocks=10,
+            metric_last_update=now,
+            publish_end=datetime.utcnow() + timedelta(days=1))
```
pmac: NIT: I'm sure it's just copypasta, but you're using the `now` variable created at the top of the method in some places and calling `datetime.utcnow()` again in others.

glogiotatidis: Thanks!

glogiotatidis:
> This looks good to me. Are you concerned that this will disable snippets even when the issue with metrics might be between telemetry and snippets service, and telemetry is in fact being collected just fine? Perhaps this state needs an alert of some kind?

Good point. This is handled in [fetch_metrics](https://github.com/mozmeao/snippets-service/blob/master/snippets/base/management/commands/fetch_metrics.py#L115) command: When there are jobs to fetch data for and we don't collect for any of them we raise `CommandError`. This will trigger the DMS which monitors the job.

Does this answer your concern @pmac?

pmac:
Yup. Sounds good. :shipit:

glogiotatidis:
Thanks!

PR #1210, opened 2019-10-29
[Issue #1112] Redirect all requests to CDN / S3

glogiotatidis:
Redirect all requests to CDN / S3 without checking first if the bundle exists on
S3.

CloudFront is configured to return a custom page (which is an empty JSON file)
when origins return 403 (Permission Denied). S3 bucket will return 403 -and not
404- when you request a file that doesn't exist because it disallows directory
listing.

This way we can return valid 404s from the Django app when needed and empty
bundles when we don't have a snippets for a channel / locale / distribution
combination, without spending time on I/O between the Django App and S3 or
hitting the cache too many times.

### DIFF
--- a/snippets/base/tests/test_views.py
+++ b/snippets/base/tests/test_views.py
@@ -209,13 +209,6 @@ class FetchSnippetPregenBundleTests(TestCase):
         )
         self.assertEqual(response.url, expected_url)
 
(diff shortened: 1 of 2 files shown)


glogiotatidis:
Thanks!

PR #1214, opened 2019-10-30
[Fix #1213] Default to `default` dist if client is not part of an experiment.

### LINKED ISSUE #1213: Default to `default` distribution if client is not part of an experiment.
We want all Firefox clients to get the default bundle for the locale / channel combination, unless they are part of an experiment when using pregen bundles in https://github.com/mozmeao/snippets-service/blob/master/snippets/base/views.py#L63

This is because non-Mozilla distributors of Firefox (e.g. Linux Distributions) override the distribution field with their identification. 

We 'll only use client's distribution if it starts with `experiment-`

glogiotatidis:


glogiotatidis:
This is a small change with tests, self merging

PR #1215, opened 2019-10-30
Add DailyJobMetrics

glogiotatidis:
This still needs tests but it appears to be working fine on local runs.

Posting here for a code review while I'm writing the tests, to speed up delivery


snippets/base/management/commands/fetch_daily_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                data_fetched = True
+
+            if not any([impressions, blocks, clicks]):
+                # Job has no metrics for date, ignore
+                continue
+
+            # Use update to avoid triggering Django signals and updating Job's
+            # and ASRSnippet's modified date.
+            DailyJobMetrics.objects.create(
```
jgmize: I think we may want to use `update_or_create` here to avoid duplicates that could result from running this command more than once for the same day.

snippets/base/management/commands/fetch_daily_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+
+        data_fetched = False
+        for job in jobs:
+            bind_data = {
+                'start_date': date.strftime('%Y%m%d'),
+                'end_date': date.strftime('%Y%m%d'),
+                'message_id': job.id,
+            }
+            try:
+                result = redash.query(settings.REDASH_QUERY_ID, bind_data)
```
jgmize: I'm wondering about the performance implications of doing a separate redash query for each job. For the small number of jobs published each day this probably doesn't matter, but for loading all of the data going back to when we started publish jobs we might want to use a different approach. We can follow up with a separate script for that if needed though.

snippets/base/management/commands/fetch_daily_metrics.py
glogiotatidis: We shouldn't be running this command more than once per day but it shouldn't hurt to add update_or_create here.

snippets/base/management/commands/fetch_daily_metrics.py
glogiotatidis: We're going to load historical data starting from Q4, since we have incomplete data for Sept and Q3 anyway. I constructed the command in this way so we can import which day separately. With a loop we will fetch all data within 24 hours which is ok for a one-off.

> For the small number of jobs published each day

Each day we need to fetch data for more than 1000 Jobs and we need about 90 minutes to do so.

glogiotatidis: Actually what @jgmize proposed was way smarter and implemented things that way. Queries only take a few seconds per day now.

glogiotatidis:
Also added a cron job to fetch daily.

Ready for review @jgmize

glogiotatidis:
Thanks!

snippets/base/admin/adminmodels.py
```diff
@@ -1019,7 +1019,7 @@ class JobAdmin(admin.ModelAdmin):
     ]
     fieldsets = [
         ('ID', {
-            'fields': ('id', 'job_status', 'snippet_name_linked', 'creator')
+            'fields': ('id', ('job_status', 'completed_on'), 'snippet_name_linked', 'creator')
```
jgmize: I missed this while reviewing: https://github.com/mozmeao/snippets-service/issues/1220

PR #1254, opened 2019-12-05
[Fix #1253] Show populated inline Template to users with view-only perms.

### LINKED ISSUE #1253: Users with view-only permissions cannot see the populated inline template

glogiotatidis:
Work around for the custom templates dropdown and the view-only permissions of the django admin.


glogiotatidis:
Thanks!

PR #1256, opened 2019-12-05
[Issue #1264] Add additional daily metrics to admin

jgmize:
Following up on https://github.com/mozmeao/snippets-service/pull/1255#pullrequestreview-327603389


jgmize:
Thanks @glogiotatidis, good catch! I've added the models to the admin init, but what do you mean by "place them in https://github.com/mozmeao/snippets-service"?

glogiotatidis:
> Thanks @glogiotatidis, good catch! I've added the models to the admin init, but what do you mean by "place them in https://github.com/mozmeao/snippets-service"?

copy paste fail, I meant https://github.com/mozmeao/snippets-service/blob/master/snippets/settings.py#L322

jgmize:
Thanks for your patience @glogiotatidis; it's been a while since I've added models to a django admin. I can see that you've put a lot of quality work into this one.

I think the latest commit should address the changes and a typo caught by the CI.

snippets/base/admin/adminmodels.py
```diff
@@ -1275,3 +1275,78 @@ def has_change_permission(self, request, obj=None):
 
     def has_delete_permission(self, request, obj=None):
         return False
+
+
+class DailySnippetsMetrics(admin.ModelAdmin):
```
glogiotatidis: Typo, this should be `DailySnippetMetrics` since everything else is also singular. This inconsistency made to make an error in `ADMIN_REORDER` and the model should be of course `DailySnippetMetrics` too

glogiotatidis:
Fixed the typo to help push this :fast_forward:

jgmize:
Thanks so much @glogiotatidis!

PR #1260, opened 2019-12-08
Add Jupyter 

jgmize:
- https://gitlab.com/mozmeao/snippets-service/-/jobs/372926034


docker-compose.yml
```diff
@@ -29,4 +46,4 @@ services:
       - CSP_REPORT_ENABLE=False
```
glogiotatidis: `web` has a bunch of env variables set originally and you only leave it with 3. Copy / paste erorr?

.gitlab-ci.yml
```diff
@@ -17,6 +17,17 @@ build-images:
     - docker build -t ${DOCKER_IMAGE_TAG} --build-arg GIT_SHA=${CI_COMMIT_SHA} --pull .
     - docker push ${DOCKER_IMAGE_TAG}
 
+build-jupyter:
```
glogiotatidis: This  section seems disconnected from the rest of the file. It does things in a different way, e.g. using compose, not tagging the built images, not using the set variables.

docker-compose.yml
jgmize: good catch, thanks! I've restored the accidentally deleted env vars in ad0ff18

.gitlab-ci.yml
jgmize: Good points, I've updated it in c09c0d1 to be more in alignment with the rest of the file.

.gitlab-ci.yml
```diff
@@ -16,6 +17,8 @@ build-images:
   script:
     - docker build -t ${DOCKER_IMAGE_TAG} --build-arg GIT_SHA=${CI_COMMIT_SHA} --pull .
     - docker push ${DOCKER_IMAGE_TAG}
+    - docker build -t ${JUPYTER_IMAGE_TAG} --build-arg DEVELOPMENT=true .
```
glogiotatidis: Let's keep this is separate step

docker-compose.yml
glogiotatidis: Thanks for updating. I belive `ENABLE_ADMIN` and the `CSP` variables are not needed to run Jupiter. Let's try to keep things tidy!

jgmize: thanks, removed in b64932c

Issue #1248, opened 2019-12-04
Fetch snippets data from the new BigQuery data source


glogiotatidis:
Starting from Fx62 (snippets) telemetry will use the BigQuery data source.

Let's fetch data for both data sources (current and new BigQuery). 

This applies to both
 - [x] `fetch_metrics` and
 - [x] `fetch_daily_metrics`  

It seems that (some) Beta installations are  using the BigQuery pipeline since 2019-12-04. 

 - [x] Reprocess data from 2019-12-04 after this is merged

PR #1262, opened 2019-12-10
Add snippets.base.etl helper functions for ETL jobs

jgmize:
These functions and their associated queries have all been tested locally with a local DB the jobs & snippets restored from prod.


jgmize:
I've rebased this on master after merging #1248 as we discussed, and I'm next I'm working on incorporating the ETL functions into the fetch_metrics command as well as adopting some of your other excellent ideas like linking to the new queries from the admin.

Issue #1275, opened 2019-12-18
Change job_fetch_metrics only update metrics on Freq Capped Jobs


glogiotatidis:
We want `fetch_metrics` job to only fetch data on Freq Capped Jobs in very regular intervals so we can pull fast out of bundles a Job that reached it's cap.

Make Impression / Block / Clicks columns in Job list view get their numbers from DailyJobMetrics.

![image](https://user-images.githubusercontent.com/584352/71070387-5cd33780-2183-11ea-98d7-800fb30ca0a8.png)

PR #1277, opened 2019-12-18
[Issue #1275] Get Job metrics from DailyJobMetrics model.

### LINKED ISSUE #1235: Snippet Admin: Separate out CTR and Block Rate in Jobs List View
Currently, the CTR is bundled with the clicks and the block rate is bundled with the blocks which makes it manual for us to separate them when we copy and paste into a separate sheet. Would like to have the CTR and block rate in its own columns to make it easier to read the data when we copy and paste. Thank you!

glogiotatidis:
[Fix #1235] Also separate out CTR and Block Rate in Jobs List View


glogiotatidis:
Thanks! :

PR #1286, opened 2020-01-08
[Fix #1275] Only update metrics for Freq Capped jobs.

### LINKED ISSUE #1275: Change job_fetch_metrics only update metrics on Freq Capped Jobs

glogiotatidis:


glogiotatidis:
Thanks!

PR #1287, opened 2020-01-09
[Fix #1231] Use OAuth Client IDs instead of names.

### LINKED ISSUE #1231: Use OAuth client ids instead of names in Firefox App Targeting.
List of client ids: https://docs.telemetry.mozilla.org/datasets/fxa_metrics/attribution.html#service-attribution

Related bug https://bugzilla.mozilla.org/show_bug.cgi?id=1596514#c5

 - [ ] After pushing to prod, generate JEXL for related Targets
 - [ ] ping mindy when is complete

glogiotatidis:


snippets/base/admin/fields.py
```diff
@@ -205,19 +210,27 @@ def __init__(self, **kwargs):
         self.widget = JEXLMultiWidget(widgets=[f.widget for f in self.fields])
 
     def compress(self, data_list):
+
```
pmac: Probably didn't mean to add this line?

glogiotatidis:
Thanks!

PR #1297, opened 2020-01-20
[Fix #1289] Add Frequency Capping ADR.

### LINKED ISSUE #1289: ADR on Frequency Capping

glogiotatidis:


docs/architecture/decisions/0005-frequency-capping.md
```diff
@@ -0,0 +1,110 @@
...
+errors.
+
+In co-ordination with Snippet Content Owner we decided that for cases
+where the Snippets Service cannot get the latest numbers for more than
+24 hours, Jobs with Global Frequency Capping will get canceled. The
+cancellation reason will state that the Jobs where prematurely
+terminated due to missing metrics.
+
+The cron job responsible for fetching the Data from Telemetry is
+monitored by a Dead Man's Snitch.
```
duallain: How frequently do we intend to run the job itself? Mostly thinking, how many failures in a row before we impact a particular bundle/snippet?  It's nice to get multiple attempts to pull this sort of information.

It might be good to include the current frequency in the context of the decision. (We explain the Redash job frequency, but not the Telemetry cron frequency.)

glogiotatidis: I see what you mean. Redash is the interface to access Telemetry so it's really the same thing in this document.

I reworded a few sentences to make it more clear. Let me know if that makes more sense

> Snippets Service will query Mozilla's Redash for Telemetry data every
ten minutes and will fetch current impressions, clicks, blocks for
each Job with set limits.

and

> The cron job responsible for fetching the Data from Redash is
monitored by a Dead Man's Snitch.

docs/architecture/decisions/0005-frequency-capping.md
```diff
@@ -0,0 +1,110 @@
...
+
+For example a Content Piece can be set to 1,000,000 Global Impressions
+and 1 Impression per User, thus indirectly driving 1,000,000 different
+users to this Content.
+
+This functionality has been lacking from the Snippet Service due to
+technical limitations imposed by the way metrics were collected and
+content selection was handled on the client side. The latest
+developments in Firefox Messaging Center and the Firefox Telemetry
+Pipeline unblock this capability.
```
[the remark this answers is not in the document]
glogiotatidis: Not strictly this one bug only but probably starting point https://bugzilla.mozilla.org/show_bug.cgi?id=1433214 

will add to the doc

glogiotatidis:
Thanks for your input all! :beers:

PR #1309, opened 2020-02-05
[Fix #1308] Include Release Jobs in Nightly Bundles.

### LINKED ISSUE #1308: Include Release Jobs in Nightly Bundles
We will include Release Jobs into nightly bundles with custom JEXL to always evaluate to `false`. This is a request from the Firefox Messaging team so they can monitor for regressions in the JEXL engine.

glogiotatidis:


glogiotatidis:
Thanks!

snippets/base/management/commands/generate_bundles.py
```diff
@@ -92,12 +105,21 @@ def handle(self, *args, **options):
                             default_storage.delete(filename)
                         continue
 
-                    data = [job.render() for job in bundle_jobs]
+                    data = []
+                    channel_job_ids = list(channel_jobs.values_list('id', flat=True))
+                    for job in bundle_jobs:
```
glogiotatidis: Slightly re-engineered this bit to make sure that in the -unlikely- event that a Job targets both `nightly` and `release` we render it normally and not with `always_eval_to_false=True`

glogiotatidis: Also updated test

Issue #1285, opened 2020-01-08
Stop pulling data from RedShift 


glogiotatidis:
Starting in version 72 (Jan 2020), Firefox Telemetry uses BigQuery instead of RedShift. 

We currently pull data from both data sources for frequency capping and performance reports. 

In about a year from now the usage of pre-72 versions will be limited and we will be able to remove the RedShift queries from the codebase.

- [x] Stop pulling for Freq Capped Jobs
- [x] Stop pulling Daily Data

glogiotatidis:
Apparently the Redshift pipeline was shut down on 2020-02-05. We can go ahead and remove this code

PR #1319, opened 2020-02-17
[Issue #1285] Stop pulling data from RedShift data store.

glogiotatidis:
The RedShift data pipeline has been shutdown, so no need to query Redash for that data. 
This only affects freq capped jobs that depend on the `fetch_metrics` management command. Daily metrics jobs will be updated in a future PR.


glogiotatidis:
Thanks!

PR #1327, opened 2020-02-21
ETL jobs update

glogiotatidis:
The main idea is the we collect daily data for each Job in each country and each Fx channel. We process the data to correct Telemetry pipeline errors and to filter out invalid entries. 

Then we store  aggregations of all events in the obj model and processed items under `details`. Most jobs have only `impression`, `click` and `block` events but other also have `go_to_scene2`, `dismiss`, `other_click`, `subscribe_error` and `subscribe_success` events.

`details` field is stored as JSON in PostgreSQL and we can in the future directly query this to produce metrics per channel or per country or a combination of both.

Another interesting process is that we query Telemetry for avg session time in `about:home` and `about:impressions` for each day. We store the number of sessions that lasted 4 to 15-and-above seconds in steps of one second. 

We use those numbers of calculate a more realist number of Impressions (from now on `Adjusted Impressions`) for each job (stored under `JobDailyPerformance.adj_impression`) which doesn't account impressions that lasted less than 5 seconds. 

This threshold is based on avg human read and comprehension speed for a sentence of the size of a snippet. As with JobPerformance, we store all the information is `DailyImpressions.details` so we can fine tune the threshold and re-calculate Adjusted Impressions if needed.


glogiotatidis:
Tests need to get updated. If you want to do a first pass already @pmac feel free

glogiotatidis:
@pmac hold off reviewing this. i'll probably make some big changes

glogiotatidis:
@pmac this is complete. Another unrelated test started failing on travis only - I can't reproduce locally - but I need to get some sleep before fixing it.

glogiotatidis:
> It looks really good to me. I don't know much about the redash API and what all you're pulling but I don't see any obvious code issues, and if it's working for you then +1

I'll schedule a meeting to demo to the team what i've been building and how we interact with redash.

Thanks for your time @pmac !

PR #1331, opened 2020-02-25
[Fix #1186] Rate limit public job list view.

### LINKED ISSUE #1186: Add caching &/or rate limiting to public jobs list view
Earlier today the snippets admin experienced some performance issues due to automated scanning against the public jobs list view (someone was using [sqlmap](http://sqlmap.org) to probe for SQL injection vulnerabilities). We may may want to use https://github.com/jsocol/django-ratelimit similar to basket to limit the impact of such scanning, and we could potentially add some caching to the views to reduce the DB load from a large number of requests to that public view.

glogiotatidis:


glogiotatidis:
> Looks good. My only comment is that `6/m` seems pretty restrictive even per IP, but it's worth a shot.
I think it's going to be fine for that view but I hear your concern. I'll go ahead and raise this to `10/m` and convert the config to a env variable so we can easily adjust if needed.

>  Do we have any metrics hooked up to this thing? Can we keep an eye on how many 429s we send?
Added sentry and we also have `GraphiteMiddleware` that logs to statsd (i've forgotten about this btw)

PR #1340, opened 2020-03-23
[Fix #1285] Stop pulling data from RedShift.

### LINKED ISSUE #1285: Stop pulling data from RedShift

glogiotatidis:
RedShift pipeline has been shut down and receives no new data.

### DIFF
--- a/snippets/base/tests/test_etl.py
+++ b/snippets/base/tests/test_etl.py
@@ -84,9 +84,7 @@ class TestUpdateJobMetrics(TestCase):
                     'channel': 'demo',
                     'country_code': 'GR',
                     'counts': 5,
-                }
-            ],
-            [
+                },
                 {
                     'message_id': '1000',
                     'event_context': '',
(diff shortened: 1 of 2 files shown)


snippets/base/tests/test_etl.py
```diff
@@ -235,7 +231,7 @@ def test_base(self):
 
                 ],
             ]
-            self.assertEqual(etl.update_impressions('2019-12-20'), 3)
+            self.assertEqual(etl.update_impressions('2019-12-20'), 2)
```
pmac: This line is failing in Travis

PR #1353, opened 2020-04-07
Snippet metrics extraction script.

glogiotatidis:
The script used to extract quarterly (although not limited to) metrics for Snippets. It exports CSV to be manually imported to a spreadsheet. 

@pmac r?

### DIFF
--- a/scripts/metrics.py
+++ b/scripts/metrics.py
@@ -1,0 +1,222 @@
+#####
+## Script for Snippet Metrics
+##
+## This is not meant to be fast or pretty. But it does the job and
+## it's easy to update.
+##
+##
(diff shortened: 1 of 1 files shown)


scripts/metrics.py
```diff
@@ -0,0 +1,222 @@
...
+from django.db.models import Sum
+
+from product_details import product_details
+
+from snippets.base.models import *
+
+
+def pprint(d):
+     print(
+          f"  Impressions; "
```
pmac, line 31: NIT: don't need to be f-strings here or in the first print of the next function.

glogiotatidis:
Thanks!

PR #1378, opened 2020-05-11
Remove obsolete CMR from Target.

glogiotatidis:
We moved away from using Client Match Rules to allow for pregen bundles. This PR removes related code


glogiotatidis:
Thanks!

PR #1381, opened 2020-05-19
[Issue #1380] Fix JSON key name for button entrypoint.

glogiotatidis:


glogiotatidis:
Minor fix and moar tests.

PR #1392, opened 2020-06-04
[Fix 1391] Allow disabling DistributionBundles.

glogiotatidis:


snippets/base/management/commands/generate_bundles.py
```diff
@@ -96,10 +96,10 @@ def handle(self, *args, **options):
                         Q(snippet__locale__code__contains=splitted_locale) |
                         Q(snippet__locale__code__contains=full_locale)).distinct()
 
-                    # If there 're no Published Jobs for the channel / locale /
-                    # distribution combination, delete the current bundle file if
-                    # it exists.
-                    if not bundle_jobs.exists():
+                    # If DistributionBundle is not enabled, or if there 're no
```
pmac, line 99: should be `if there are no`

glogiotatidis:
Thanks!

PR #1412, opened 2020-07-09
[Fix #1413] Fetch DailyImpressions only once per week.

### LINKED ISSUE #1413: Fetch Daily Impressions Data once per week
DailyImpressions are used to calculate an approximation of how many
Impressions are actually valid impressions, i.e. User has spend enough
time on the page to read and understand the message.

Metrics have shown that the percentage of time spend of New Tab Page is
pretty stable and around 63% of the time Users stay 5 seconds of more on
the page.

Since the number is pretty stable, we don't need to pull it every day.
Instead this PR changes the pull frequency to once per week to save on
resources.

- [x] can we set completed_on for CANCELED Jobs as well?
- [x] if so, migrate the existing data to have proper dates

glogiotatidis:
DailyImpressions are used to calculate an approximation of how many
Impressions are actually valid impressions, i.e. User has spend enough
time on the page to read and understand the message.

Metrics have shown that the percentage of time spend of New Tab Page is
pretty stable and around 63% of the time Users stay 5 seconds of more on
the page.

Since the number is pretty stable, we don't need to pull it every day.
Instead this PR changes the pull frequency to once per week to save on
resources.

### DIFF
--- a/snippets/base/etl.py
+++ b/snippets/base/etl.py
@@ -1,9 +1,10 @@
 import collections
 import json
-
+from datetime import datetime, timedelta
 from urllib.parse import urlencode
 
 from django.conf import settings
+from django.db.models import Q
 from django.db.transaction import atomic
 from redash_dynamic_query import RedashDynamicQuery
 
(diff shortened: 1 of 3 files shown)


glogiotatidis:
> Looks good to me. No need for updating tests?

Good points. Current tests don't tough this (yet).

PR #1420, opened 2020-07-27
[Fix #1419] Set entrypoint_* only to about:logins

### LINKED ISSUE #1419: Append entrypoint url argument only to `about:logins` page

glogiotatidis:


glogiotatidis:
Minor fix, tests passed

PR #1449, opened 2020-10-14
[Fix #1447] Allow deletion of Icons not currently in use.

### LINKED ISSUE #1447: Allow deletion of Icons not currently in use
Hi @glogiotatidis & @pmclanahan 

We have a lot of icons no longer appliable due to their pixel size or interface incompatability. Unable to remove them without deleting unpublished snippets (no current job tied to snippet). However, we would like to retain the snippets for reference or reuse in the future. Recommend icon archival button that allows the user to categorize and hide old icons as we continue to grow/change. Once an icon is removed to archive, the snippet that retains that icon should prompt a change to a new icon when it comes time to publish a job or the icon is stripped or a generic icon indicator to be updated. Implimentation is needed as we have new publishers joining the platform, worried about governance when it comes to programming. Thanks!

Screen Shot 2020-09-30 at 10 29 33 AM

glogiotatidis:


pmac (approved):
💯

glogiotatidis:
Thanks!

Issue #1464, opened 2020-11-04
Use the same code to pregenerate and generate on the fly for dev  


glogiotatidis:
We currently have two bundle generation implementations:
 - The current offline bundles generator,  that creates and saves the bundles to S3 and
 - The previous implementation that was creating bundles on the fly upon HTTP request.

Offline bundle generation is working fine the last year. Drop the legacy implementation and provide a way for on-the-fly bundles for development purposes.

## Task list
- [x] Remove `USE_PREGEN_BUNDLES` from `snippets-config`

PR #1467, opened 2020-11-09
[Issue #1464] One path to bundle generation.

glogiotatidis:


glogiotatidis:
@pmac i continue the cleanup before I go ahead with new features, specifically #1446

pmac (changes requested):
Looking good. Travis is showing there's a test to fix, but other than that it's nice to see so much debt deletion!!

glogiotatidis:
> that it's nice to see so much debt deletion!!

aaaand i've more coming up! Thanks!

Issue #1468, opened 2020-11-09
Simplify snippets-prod deployments


glogiotatidis:
Replace snippets prod deployments with minimal python apps that redirect to the correct pregen bundle. 

## Task list
 - [x] Deploy to frankfurt and oregon
 - [x] Cleanup snippets-config snippets-prod
 - [x] Evaluate frankfurt RDS replica removal
 - [ ] Eval Lambda failovers

PR #1476, opened 2020-11-16
[Issue #1468] Simple redirector app.

glogiotatidis:
I need to update `gitlab-ci.yml` so the pipeline deploys the redirector but I'll do that after this is merged.


pmac (approved):
Looks good to me. I built the new service and ran it and it worked well locally. We'll probably want a way to start up gunicorn with more than one worker or to look into using gevent or something for perf, but it's a great start.

glogiotatidis:
> Looks good to me. I built the new service and ran it and it worked well locally. We'll probably want a way to start up gunicorn with more than one worker or to look into using gevent or something for perf, but it's a great start.

very good point, added https://github.com/mozmeao/snippets-service/pull/1476/files#diff-ce1fb1f4ed3fde75821023f737f5e4aad410e3e5e4dc878f384225793b02e44eR51

PR #1484, opened 2020-12-01
December python updates

glogiotatidis:


glogiotatidis:
Tests pass here https://gitlab.com/mozmeao/snippets-service/-/pipelines/223572616

PR #1485, opened 2020-12-02
Add X-Backend-Server header in redirector.

glogiotatidis:


glogiotatidis:
Had to merge this to complete the new cluster setup, still interested in your review pmac.

pmac (commented):
Are any of those env vars auto popullated by k8s? That would be cool. Looks good to me either way. The only other thing I thought of was that this could have been implemented as a WSGI middleware to avoid having to decorate every view, but this is so simple that it's probably overkill to do that. ⛄

glogiotatidis:
> Are any of those env vars auto popullated by k8s? That would be cool. Looks good to me either way. The only other thing I thought of was that this could have been implemented as a WSGI middleware to avoid having to decorate every view, but this is so simple that it's probably overkill to do that. snowman

yes we populate  K8S_NAMESPACE and K8S_POD_NAME  from the cluster but it needs config so it's semi-automatic.

PR #1487, opened 2020-12-07
[Fix #1486] Fix Fundraising ETL for clicks.

### LINKED ISSUE #1486: Fundraising click metrics are not collected in snippets-admin
It seems that Fundraising clicks follow a different pattern in telemetry and thus they don't get collected by snippets-admin etl.

glogiotatidis:


glogiotatidis:
This is a small change with tests. Will merge to fix the data in prod today

PR #1565, opened 2022-01-24
Remove ipdb from dev requirements, to avoid including iPython

stevejalim:
Remedies PR #1564 

Neither `ipdb` nor `ipython` are needed for an actual dev build, they are more local tooling

Reqs were updated via `pip-tools` in the docker container. Looks like some other deps were updated while removing `ipdb` and its subdeps


redirector/requirements.txt
```diff
@@ -4,14 +4,14 @@
...
-    --hash=sha256:31b2eced602aa8423c2aea9c76a724617ed67cf9513173fd3a4f03e3a929c7e6 \
-    --hash=sha256:832aa3cde19744e49938b91fea06d69ecb9e649c93ba974535d08ad92164f700
+attrs==21.4.0 \
+    --hash=sha256:2d27e3784d7a565d36ab851fe94887c5eccd6a463168875832a1be79c82828b4 \
+    --hash=sha256:626ba8234211db98e869df76230a137c4c40a12d72445c45d5f5b716f076e2fd \
     # via pytest
 bottle==0.12.19 \
     --hash=sha256:a9d73ffcbc6a1345ca2d7949638db46349f5b2b77dac65d6494d45c23628da2c \
-    --hash=sha256:f6b8a34fe9aa406f9813c02990db72ca69ce6a158b5b156d2c41f345016a723d
-    # via -r redirector/requirements.in
```
stevejalim, line 14: Not sure why pip-tools is removing the `redirector` prefix here, but the right file is getting the right dep based on redirector/requirements.in

robhudson (approved):
Looks fine on the surface

PR #1582, opened 2022-04-11
Upgrade Django and django-jinja

stevejalim:
Django 2.2.28 is a security release, while upgrading django-jinja was needed
to allow the service to build/run, rather than blow up.

Given the horizon for snippets, I think this is a relatively safe/low risk
upgrade of django-jinja


robhudson (approved):
lgtm

